Issue #72, opened 2015-11-29
Add flag to skip git init + git add


kallisti5:
```
13:22 <@PulkoMandy> we could probably run haikuporter without git on the slaves                                                                                                                             
13:22 <@PulkoMandy> (using plain extract + patch)                                                                                                                                                           
13:22 <@PulkoMandy> the git is only useful for --extract-patchset      
```

kallisti5:
Some large (LARGE) projects, the git add takes a long, long time. (webkit 5+ hours).  haikuporter could use a flag to skip the git repo creation process.

jessicah:
Looking at the pull request, this would only work if there are no patches to apply, which very much limits the utility of this option.

Shouldn't it instead apply patches with patch instead of git am?

KapiX:
OK, `patch` does work with git patchsets. I thought it does not. @DarkmatterVale can you rework your solution?

DarkmatterVale:
@KapiX Yes, I am trying to find a solution right now.

KapiX:
Fixed by 7e5661a994db976ae897e2ef42b07382cb1699b2.

PR #73, opened 2016-01-08
Add option to not create a git repo

DarkmatterVale:
This PR adds an option to ignore creating a git repo. This drastically reduces the time required to download & prepare packages.

It has a built in safety feature which will make the program create a repo if patches exist. This is used to prevent any issues with patches attempting to be applied when no git repo is available.

### DIFF
--- a/HaikuPorter/Options.py
+++ b/HaikuPorter/Options.py
@@ -86,9 +86,9 @@ def parseOptions():
 	parser.add_option('-e', '--extract-patchset',
 					  action='store_true', dest='extractPatchset', default=False,
 					  help='extract current patchset(s) from port source(s)')
-	parser.add_option('-G', '--init-git',
-					  action='store_true', dest='initGitRepo', default=False,
-					  help='create git-repo(s) for port source(s)')
+	parser.add_option('-G', '--no-git-repo',
+					  action='store_true', dest='noGitRepo', default=False,
+					  help="don't create git-repo(s) for port source(s)")
 	parser.add_option('-B', '--patch-files-only',
 					  action='store_true', dest='patchFilesOnly',
 					  default=False,
--- a/HaikuPorter/Source.py
+++ b/HaikuPorter/Source.py
@@ -262,6 +262,11 @@ class Source(object):
 
 		# use an implicit git repository for improved patch handling.
 		ensureCommandIsAvailable('git')
+		
+		# Check to see if there are any patches that need to be applied.
+		if not self.patches and getOption('noGitRepo'):
+			return False
+
 		if not self._isInGitWorkingDirectory(self.sourceDir):
 			# import sources into pristine git repository
 			self._initImplicitGitRepo()


DarkmatterVale:
This solves #72

HaikuPorter/Port.py
```diff
@@ -102,6 +102,8 @@ def __init__(self, name, version, category, baseDir, outputDir,
 		self.repositoryDir = repositoryDir
 		self.recipeIsBroken = False
 		self.recipeHasBeenParsed = False
+		self.initGitRepo = initGitRepo
```
waddlesplash: If this isn't actually used anywhere, you can nuke it.

HaikuPorter/Options.py
```diff
@@ -80,6 +80,9 @@ def parseOptions():
 					  action='store_true', dest='debug', default=False,
 					  help="show Python stack traces for fatal errors")
 
+	parser.add_option('--no-git-repo',
```
waddlesplash: Maybe it'd be best to invert this (`--git-repo`) and default it to `True`?

DarkmatterVale: Sure, I will swap it right now.

DarkmatterVale:
I want to squash all commits down to one, correct?

waddlesplash:
Yes.

waddlesplash:
LGTM; I'd wish for someone more skilled in python than myself to review but IDK who that would be...

DarkmatterVale:
I'm skilled in Python, but I don't think I can review my own PR :)

HaikuPorter/Port.py
DarkmatterVale: As requested in IRC, I removed the other variable and am using this one now.

waddlesplash:
@mmlr, do you know enough Python to review this?

HaikuPorter/Source.py
```diff
@@ -260,6 +261,17 @@ def patch(self, port):
 			print 'Skipping patchset for ' + self.fetchTargetName
 			return True
 
+		# Check to see if there are any patches that need to be applied
+		if not self.patches:
+			# if init of a repo
```
KapiX: Superfluous comment.

HaikuPorter/Source.py
```diff
@@ -260,6 +261,17 @@ def patch(self, port):
...
+			if self.initGitRepo:
+				# init repo
```
KapiX: Same here.

HaikuPorter/Source.py
```diff
@@ -260,6 +261,17 @@ def patch(self, port):
...
+				self._initImplicitGitRepo()
+			
+			# return since no patches need to be applied
+			return False
+			
+
 		# use an implicit git repository for improved patch handling.
 		ensureCommandIsAvailable('git')
```
KapiX: This should probably go up.

HaikuPorter/Source.py
```diff
@@ -29,13 +29,14 @@
...
 	def __init__(self, port, index, uris, fetchTargetName, checksum,
-				 sourceDir, patches, additionalFiles):
+				 sourceDir, patches, additionalFiles, initGitRepo):
 		self.index = index
 		self.uris = uris
 		self.fetchTargetName = fetchTargetName
 		self.checksum = checksum
 		self.patches = patches
 		self.additionalFiles = additionalFiles
+		self.initGitRepo = initGitRepo
```
KapiX: You can grab this from `port` variable, so additional argument is not needed.

HaikuPorter/Options.py
korli: Am 08.01.2016 8:37 nachm. schrieb "waddlesplash" <notifications@github.com>:
>
> In HaikuPorter/Options.py:
>
> > @@ -80,6 +80,9 @@ def parseOptions():
> >    action='store_true', dest='debug', default=False,
> >    help="show Python stack traces for fatal errors")
> >
> > + parser.add_option('--no-git-repo',
>
> Maybe it'd be best to invert this (--git-repo) and default it to True?

I thought this option would only be interesting for buildbots. Now what?

KapiX:
`--init-git` option is disabled by default. This is changing current default behavior and can cause frustration if not noticed (for example when trying to extract patches from changed source). Please enable it.

DarkmatterVale:
How do I set it disabled? Should I change the var name then?

KapiX:
https://github.com/DarkmatterVale/haikuporter/blob/master/HaikuPorter/Options.py#L89
Yes, you need to change the name to reflect the functionality. Since this option was not yet used in the code, `-G` can be left as it is.

mmlr:
To reduce the clutter due to passing the option around I'd suggest using getOption('initGitRepo') in the one place where it is needed. Also I'd suggest naming the command line parameter '--no-implicit-git-repository' or at least '--no-git' instead of '--nogit' as it's clearly two words.

KapiX:
@mmlr as are '--nopatch', '--nobuild', '--nochroot' etc. Is there any reason they do not have hyphen between the words?

mmlr:
Yes, the existing inconsistent ones probably need to stay around for compatibility. Still I'd personally avoid introducing new ones that we then need to carry forward.

DarkmatterVale:
Good to know for future contributions. Thank you @mmlr !!

Issue #70, opened 2015-10-18
extractDebugInfo should preserve resources


humdingerb:
As explained by Ingo in this [commit comment](https://github.com/haikuports/haikuports/commit/7fce22cf69aed95bd8f73172c0d61d302793a07e#commitcomment-13784509) of Sequitur. Excerpts from it:
> The `defineDebugInfoPackage` shell function (defined in ShellScriptlets.py) records the paths of the files whose debug info shall be extracted. After the `INSTALL` function of the recipe is finished, `packageDebugInfos` is called which in turn calls `extractDebugInfo` for each recorded file path. `extractDebugInfo` is where the extraction of the debug info happens and where the original files are stripped afterwards.

and:

> For the time being `extractDebugInfo` should be adjusted to first check whether the file has resources and, if so, save them before stripping and re-add them afterwards.

hannahyp:
I'd like to fix it. How can I tell if a file has resources?

pulkomandy:
In C++, you would use the BResource class, but in this case you will have to write a shell script.
Maybe the "resattr" tool could be used to extract resources and save them to a temporary file.

hannahyp:
Tried resattr but it didn't seem to work because it doesn't extract resource types like APPV and BBMP. Also tried other res and elf commands, but none of them worked.

mmuman:
There is 'listres'. And 'xres' should be able to copy them all.

hannahyp:
'listres' only lists the types. I can't get it to extract anything. 'xres' extracts everything including non-resource data, so that when I add the extracted .rsrc file back to the binary, it has the same size as its non-stripped version.

I've written a Python function, but it doesn't run in ShellScriplets.py. How can I add Python code to the shell script?

pulkomandy:
This sounds strange, because xres should extract only resource data (and not debug information, which is part of the ELF file itself). Maybe the file(s) you tried this on had lot of resources and not that much debug info?

hannahyp:
I tried with both Pe and PecoRename, and both had the same result. I am writing a shell script to doing it the hard way. I hope to push the code soon.

hannahyp:
To recreate the result I got:

cp Pe Pe.ori
strip -S Pe
ls -l Pe.ori Pe
xres -o Pe.rsrc Pe.ori
xres -o Pe Pe.rsrc
ls -l Pe.ori Pe

Pe.ori and Pe have the exactly the same size now.

pulkomandy:
[...]

PR #75, opened 2016-01-17
extractDebugInfo should preserve resources #70

hannahyp:


HaikuPorter/ShellScriptlets.py
```diff
@@ -676,8 +676,18 @@ def getShellVariableSetters(shellVariables):
 
 	local objcopy=$(getTargetArchitectureCommand objcopy)
 	local strip=$(getTargetArchitectureCommand strip)
+	local xres=$(getTargetArchitectureCommand xres)
```
pulkomandy: You don't need this because there isn't an xres-x86 (the same binary works for all architectures directly).

HaikuPorter/ShellScriptlets.py
```diff
@@ -676,8 +676,18 @@ def getShellVariableSetters(shellVariables):
...
+	local tmpfile="tmp.rsrc"
```
pulkomandy: It seems a bad idea to always use the same filename, I would recommend using the mktemp command to generate a random (unique) file for you instead. This could avoid problems in parallel builds.
Aternatively, you can use the original filename and append .rsrc.

hannahyp:
Thanks for the advice! I didn't know about mktemp but shouldn't have made the xres mistake as it's already used directly in the code.

Issue #47, opened 2015-06-30
Haikuporter not generating dependencies for all older versions (which hence can't be built)


soakbot:
(Original issue 47 created by ciand on 2014-12-04T21:05:42.288201+00:00)

I'm looking at an issue with libffi and wanted to build the previous version which still has a fully functional .recipe present - but the .dependencyinfo file hadn't been generated for it. On deleting the repository folder to force a full rebuild I noticed that it was skipped without giving a reason.

Other ports with multiple versions have the .dependencyinfo file generated for all recipes

PR #76, opened 2016-01-21
Fix support for building older versions of packages

### LINKED ISSUE #69: HP doesn't create the DependencyInfo for the requested recipe version
As example, look at cdrtools. There are recipes for old alpha versions (~a27 to ~a31) and the final version:
> cdrtools-3.01.recipe
> cdrtools-3.01~a27.recipe
> cdrtools-3.01~a29.recipe
> cdrtools-3.01~a30.recipe
> cdrtools-3.01~a31.recipe

Doing a "haikuporter cdrtools" will build the last alpha (cdrtools-3.01~a31).
Doing a "hakuporter cdrtools-3.01" tries to build the final, but stops with
```
[Errno -2147459069] No such file or directory:
'/HiQ-Data/sources/haikuports/repository/cdrtools-3.01.DependencyInfo'
```
Looking into the repository folder, I see that the DependencyInfo files for cdrtools-3.01~a31 are generated instead. The same happens if I try to build any other version of cdrtools.

Related to #47 ?

DarkmatterVale:
This closes #69 and #47.

Turns out the most active version of each port was being used including the port the user would like to build. This causes the wrong dependency file to be generated.

### DIFF
--- a/HaikuPorter/Main.py
+++ b/HaikuPorter/Main.py
@@ -375,6 +375,9 @@ class Main(object):
 
 		allPorts = self.repository.allPorts
 
+		# make sure the correct dependencyInfo-file has been created
+		self.repository.supportBackwardsCompatibility(port.name, port.version)
+
 		# HPKGs are usually written into the 'packages' directory, but when
 		# an obsolete port (one that's not in the repository) is being built,
 		# its packages are stored into the .obsolete subfolder of the packages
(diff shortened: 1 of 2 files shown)


DarkmatterVale:
Does anyone have any comments on this PR or can we get it merged?

waddlesplash:
I don't feel that confident in my knowledge of HaikuPorter internals to really approve this, so it someone else who knows better than I do would care to review it, that'd be best.

Otherwise I'll just merge it.

PR #80, opened 2016-01-25
Fix error where --get doesn't work

### LINKED ISSUE #68: haikuporter --get cannot initialize the source tree.
Checkout of ports tree not done yet, I wanted to use --get for that purpose, but it fails:
```
/Dev/Haiku/hporter# ./haikuporter --get
Error: The version of the recipe file format used in the ports tree is older than the one supported by haikuporter.
Error: Please upgrade the ports tree.
```

Expected: haikuporter should be able to create the ports tree itself in that case.

DarkmatterVale:
This closes #68.

Turns out the _checkFormatVersions() was being called before _updatePortsTree(), which would cause an issue if the haikuports repo was not installed. In addition, the URLs to the repos were not correct and threw errors.

### DIFF
--- a/HaikuPorter/Main.py
+++ b/HaikuPorter/Main.py
@@ -56,6 +56,11 @@ class Main(object):
 		self.treePath = Configuration.getTreePath()
 		self.outputDirectory = Configuration.getOutputDirectory()
 
+		# if requested, checkout or update ports tree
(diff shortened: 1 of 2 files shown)


waddlesplash:
Thanks!

PR #81, opened 2016-02-13
Add support for .7z archives

DarkmatterVale:
I _think_ it works :D

### DIFF
--- a/HaikuPorter/Utils.py
+++ b/HaikuPorter/Utils.py
@@ -14,6 +14,7 @@ from subprocess import PIPE, Popen, CalledProcessError
 import sys
 import tarfile
 import zipfile
+import py7zlib
 
 if sys.stdout.isatty():
 	colorWarning = '\033[1;36m'
@@ -84,6 +85,7 @@ def unpackArchive(archiveFile, targetBaseDir, subdir):
 
 	if subdir and not subdir.endswith('/'):
 		subdir += '/'
+	extension = archiveFile.split('/')[-1].split('.')[-1]
 	# unpack source archive
 	if tarfile.is_tarfile(archiveFile):
 		tarFile = tarfile.open(archiveFile, 'r', tarinfo=MyTarInfo)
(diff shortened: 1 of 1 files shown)


waddlesplash:
Can you try it? :p

DarkmatterVale:
I did, using http://7-zip.org/a/7z1514-src.7z but I get the output "nothing to commit". I think this is an issue with haikuporter, and not the extraction (because haikuporter appears to have extracted the files fine).

waddlesplash:
Sooo can you investigate it?

KapiX:
I think you might be more successful if you unpack the files into `targetBaseDir` instead of `self.workDir`.
Also, use `subdir` like the other unpack methods do.

DarkmatterVale:
I will make the changes. Thank you for the review!!

DarkmatterVale:
Hmmm...So I am having an issue. The log output (when I run a simple test recipe) is:

```
~> hp test_7z
Warning: haiku-apps/tipster-1.0.0        is overruled by duplicate in haiku-apps - please remove one of them
Checking if any package-infos need to be updated ...
Looking for stale dependency-infos ...
Checking if package dependencies need to be updated ...
        updating dependency infos of test_7z-1.0.0
----------------------------------------------------------------------
haiku-apps::test_7z-1.0.0
        /boot/home/haikuports/haiku-apps/test_7z/test_7z-1.0.0.recipe
----------------------------------------------------------------------
Skipping download of source for 7z1514-src.7z
Validating checksum of 7z1514-src.7z
Unpacking source of 7z1514-src.7z
OPENING FILE
FILE OPENED...CREATING ARCHIVE
invalid type '\x19'
~>
```

Any ideas why I am getting a '\x19'?

korli:
Please don't add py7zlib as a hard dependency. This will cause trouble for bootstrapping.

mmuman:
Sorry you'll have to rework this after my last commit.
How about using the 7za command instead so it won't force python to import stuff?

waddlesplash:
superseded by #142

Issue #46, opened 2015-06-30
Support tar archives with hard links


soakbot:
(Original issue 46 created by cdesai on 2014-12-04T17:41:01.291186+00:00)

Copy paste from https://bitbucket.org/haikuports/haikuports/pull-request/259/do-not-merge-update-cdrtools-to-301-a25/diff#comment-4158357

The problem with that archive (upstream source for cdrtools) has is that it contains hard links, which aren't supported by BFS.

The 'tar' utility has an option for that,

-Q, --replace-hardlinks replace hard links with corresponding symlink when extracting

but I wasn't able to find something like that for the haikuporter python implementation which uses tarfile.

The archive that I created replaces those hard links with symbolic links.

soakbot:
(Original issue 46 last updated on 2014-12-13T22:46:31.544159+00:00)

(Issue automaticaly closed due to status in Bitbucket: resolved)

soakbot:
(Original comment by zooey on 2014-12-13T22:46:31.563588+00:00)

Fixed in 20f7c9a, which adds code to haikuporter that silently converts hard links to symlinks when extracting tar archives.

PR #91, opened 2016-06-30
Call tarfile.open with dereference=True to handle hard links.

fbrosson:
* **`Utils.py`**: Use **`dereference=True`** (instead of **`tarinfo=MyTarInfo`**) when calling **`tarfile.open`** to handle archives with hard links.
* **`SourceFetcher.py`**: Use **`--hard-dereference`** when calling **`tar`**.

### DIFF
--- a/HaikuPorter/SourceFetcher.py
+++ b/HaikuPorter/SourceFetcher.py
@@ -56,10 +56,10 @@ def unpackCheckoutWithTar(checkoutDir, sourceBaseDir, sourceSubDir, foldSubDir):
 	sourceDir = sourceBaseDir + '/' + sourceSubDir \
 		if sourceSubDir else sourceBaseDir
 	if foldSubDir:
-		command = ('tar -c -C "%s" --exclude-vcs | tar -x -C "%s"'
+		command = ('tar -c -C "%s" --exclude-vcs | tar -x --hard-dereference -C "%s"'
 				   % (foldSubDir, sourceDir))
 	else:
-		command = 'tar -c --exclude-vcs . | tar -x -C "%s"' % sourceDir
+		command = 'tar -c --exclude-vcs . | tar -x --hard-dereference -C "%s"' % sourceDir
 	check_call(command, cwd=checkoutDir, shell=True)
 
 	if foldSubDir:
@@ -270,10 +270,10 @@ class SourceFetcherForGit(object):
 		sourceDir = sourceBaseDir + '/' + sourceSubDir \
 			if sourceSubDir else sourceBaseDir
 		if foldSubDir:
-			command = ('git archive %s "%s" | tar -x -C "%s"'
+			command = ('git archive %s "%s" | tar -x --hard-dereference -C "%s"'
 					   % (self.rev, foldSubDir, sourceDir))
 		else:
-			command = 'git archive %s | tar -x -C "%s"' % (self.rev, sourceDir)
+			command = 'git archive %s | tar -x --hard-dereference -C "%s"' % (self.rev, sourceDir)
 		check_call(command, shell=True, cwd=self.fetchTarget)
 
 		if foldSubDir:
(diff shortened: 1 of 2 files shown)


fbrosson:
Should I also revert the rest of the commit mentionned in https://github.com/haikuports/haikuporter/issues/46#issuecomment-116915237 by dropping the bloc that defines **`class MyTarInfo`**?

waddlesplash:
I'm too uncertain of this code to merge this, sorry.

fbrosson:
Works fine for me.
With the current haikuporter, when you unpack e.g. cdrtools-3.02~a06 (which has a few hard links) you get them as symlinks but they are relative to work-3.02~a06/sources/, so they are unusable. Hopefully these hard links are only readme files (and are not necessary for the compilation).
With this PR, you get copies instead of incorrect symlinks. If the conversion from hard link to symlink was correct then I wouldn't be suggesting this change. But hard links are not (currently) being correctly handled.
If you still prefer to not merge this PR, that's OK for me. I'll continue to use my locally modified haikuports.

pulkomandy:
Replacing hardlinks with copies may break things in the following situation:
- a and b are hardlinks to the same data
- compilation process modifies a
- compilation process reads b and expects changes done in previous step to be visible

Replacing them with symlinks usually avoids this problem, and is the correct way to go about this. However, the symlinks should ideally use relative paths.

The old RelativizeSymLink tracker add-on could do this (and I have the source somewhere). But it should be done the Python way in this case, I guess.

fbrosson:
OK, I got it. Let me add this, however, just for the record ;-)
The symlinks that the current version of HaikuPorter creates are not relative to the directory where they are created but to the directory where the tarball is extracted. So we cannot even expect them to point to the file they were supposed to be a hard link of. Here's what we get when we unpack cdrtools:
```
$ haikuporter -b cdrtools-3.02~a06

$ cd haikuports/app-cdr/cdrtools/work-3.02~a06/sources/cdrtools-3.02

$ ls -go BUILD COMPILE INSTALL README.compile 
lrwxrwxrwx 1    28 Jul  1 18:50 BUILD -> cdrtools-3.02/README.compile
lrwxrwxrwx 1    28 Jul  1 18:50 COMPILE -> cdrtools-3.02/README.compile
lrwxrwxrwx 1    28 Jul  1 18:50 INSTALL -> cdrtools-3.02/README.compile
-r--r--r-- 1 19660 Jul 28  2015 README.compile

$ cd cdrecord
$ ls -go COPYING LICENSE LIMITATIONS 
-r--r--r-- 1 2344 Jul  1 19:04 COPYING
lrwxrwxrwx 1   30 Jul  1 18:51 LICENSE -> cdrtools-3.02/cdrecord/COPYING
lrwxrwxrwx 1   30 Jul  1 18:51 LIMITATIONS -> cdrtools-3.02/cdrecord/COPYING
```
So, if there existed a package with hard links that were needed by the compilation, then it would probably fail to build due to the wrong relative links.
So the real reason why we should not merge this PR is that we would no longer be able to detect these cases. We would, as you have described, end up with packages built from files with incorrect content.
**I therefore agree that we should not merge this PR.** Shall I close it?

BTW, are there any kernel developers willing to add support for hard links in BFS?

pulkomandy:
hard links in BFS are not going to happen. This is not due to lazyness of our kernel programmers, it is a limitation in the design of on-disk structures for BFS making it not possible to have hard links. Fixing this would make "new" BFS incompatible with the current implementation, and possibly need major rework of the way the file system works.
We plan to replace the filesystem at some point (either with a new design or reusing an existing FS from another OS), but this probably wont happen anytime soon.

I think the way to go for this is symlinks with an absolute path. I'm not sure how we can achieve this, however.

fbrosson:
OK, thanks for the info.
Regarding the symlinks, maybe it is possible to teach MyTarInfo how to fix the path of the target when it converts hard links to symlinks, but I wouln't know how to do that, so I'll close this PR for now.

PR #104, opened 2016-12-12
Add support for 'reconnect' state in builder and build master.

netpositive36:


HaikuPorter/BuildMaster.py
```diff
@@ -460,8 +467,7 @@ def _makePackageVisible(self, packageName):
 	def status(self):
 		return {
 			'name': self.name,
-			'lost': self.lost,
-			'available': self.available,
+			'status': self.state,
```
korli: Why not 'state'?

netpositive36:
Updated.

Issue #109, opened 2017-01-18
Add support for using the same source tree on different arches


diversys:
I have haikuports folder on my /data partition. Normally I'm using gcc2h install but when I boot to x86_64 and try to build something from /data/haikuports haikuporter is updating dependencies for each and every recipe.

pulkomandy:
We should fix this by having a "packages" dir for each architecture out of the box. Another variable could be added to haikuports.conf to allow setting the package directory, for example.

But, there is also another problem: the "work" directories can also be mixed between different architectures. We should also move these so they have different locations depending on the architecture being built.

fbrosson:
Sure. I tried to create such a variable some time ago, but did not succeed, so I switched to the easier workaround.
That said, I do agree that the best thing to do is fix this the way you are suggesting.

BTW, would it be acceptable to keep all work-\* directories in the main source tree, but renaming them to include the name of the primary arch, that is, using:
* work-$targetArchitecture-$portVersion for primary arch builds
* work-$targetArchitecture-$effectiveTargetArchitecture-$portVersion for secondary arch builds

instead of what we have now:
* work-$portVersion for primary arch builds
* work-$effectiveTargetArchitecture-$portVersion for secondary arch builds

?

Hmm, instead of renaming the work-\* subdirectories we could keep them with their current names but define a flag.arch-$targetArchitecture file which haikuporter would:
* create when it populates the work-\* subdirectory
* delete when haikuporter -c is used.
* check for its presence and stop if it sees wrapper-script but not the flag.arch-$targetArchitecture.

pulkomandy:
I would find it nice to have the work dir live side by side (no need to clean anymore when I build webkit for 32 then 64 bit on the same machine).

Adding more stuff to the work-* dir names will hurt tab completion further, so I lean towards moving them elsewhere or adding an extra directory level. Maybe work-$portVersion/$targetArchitecture$secondaryArchSuffix or something similar?

korli:
@diversys you have the possibility to define a different OUTPUT_DIRECTORY in each haikuports.conf. This is  also great because the resulting $OUTPUT_DIRECTORY/packages directory only contains hpkgs built for the relevant architecture. Hope this helps.

diversys:
Turns out it is even documented at https://github.com/haikuports/haikuports/wiki/HaikuPorter-HaikuPortsConf

PR #110, opened 2017-01-18
Include buildPlatform.targetArchitecture + '-' in workDir

fbrosson:
so that each architecture has a unique work-* directory even if the haikuports tree is shared with multiple architectures.

Also put $portVersion right after 'work-' rather than at the end.
    
The new names for the work directories will be:
* "work-$portVersion-$targetArch" if building for a primary arch,
* "work-$portVersion-$targetArch-$secondaryArch" otherwise.

NOTICE: You can run: **`rm -rf /path/to/haikuporter/*/*/work-*`**

before or after you pull this commit, in order to get rid of the old work-\* directories. This is recommended, especially if you use x86_gcc2 hybrid or x86 hybrid or plan to do so.

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -112,11 +112,17 @@ class Port(object):
 		self.filter = None
 
 		if secondaryArchitecture:
-			self.workDir = (outputDir + '/work-' + secondaryArchitecture
-							+ '-' + self.version)
+			self.workDir = (outputDir + '/work-'
+				+ self.version + '-'
+				+ buildPlatform.targetArchitecture + '-'
+				+ secondaryArchitecture
+			)
 			self.effectiveTargetArchitecture = self.secondaryArchitecture
 		else:
-			self.workDir = outputDir + '/work-' + self.version
+			self.workDir = (outputDir + '/work-'
+				+ self.version + '-'
+				+ buildPlatform.targetArchitecture
+			)
 			self.effectiveTargetArchitecture = buildPlatform.targetArchitecture
 
 		recipeCacheDir = os.path.join(self.repositoryDir, 'recipeCache')
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -113,14 +113,14 @@ class Port(object):
 
 		if secondaryArchitecture:
 			self.workDir = (outputDir + '/work-'
-				+ self.version + '-'
+				+ self.version + '/'
 				+ buildPlatform.targetArchitecture + '-'
 				+ secondaryArchitecture
 			)
 			self.effectiveTargetArchitecture = self.secondaryArchitecture
 		else:
 			self.workDir = (outputDir + '/work-'
-				+ self.version + '-'
+				+ self.version + '/'
 				+ buildPlatform.targetArchitecture
 			)
 			self.effectiveTargetArchitecture = buildPlatform.targetArchitecture
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -863,7 +863,7 @@ checkedUnmount()
 trap '' SIGINT
 
 # try to make sure we really are in a work directory
-if ! echo $(basename $PWD) | grep -qE '^work-'; then
+if ! echo $(basename `dirname "$PWD"`) | grep -qE '^work-'; then
 	echo "cleanupChroot invoked in $PWD, which doesn't seem to be a work dir!"
 	exit 1
 fi


fbrosson:
This is to address one of the issues discussed in #109.
What do you think?

It works fine for me.
The only problem is that we will probably have to clean all work-\* directories (with haikuporter -c) **_before_** updating haikuporter.
Of course, we can also remove them after with rm.

waddlesplash:
Looks good to me.

fbrosson:
Thanks @waddlesplash ;-)
@pulkomandy, @korli, @waddlesplash, should we teach haikuporter how to do a "-c" on all packages when no args are supplied (in order to invite every user to call haikuporter -c before updating their haikuporter)?
Or would it be acceptable to just warn the users that they need to do a
```
find /path/to/haikuports -type d -name "work-*" -prune -exec rm -rf {} \;
```
Or would it be OK to say: "after this update, you should delete your work-*" subdirectories in any directory where you had previously built a package.but did not clean its work-\* directory" ?

BTW, I wanted to do this test:
1) choose a package which builds on all archies (e.g. libtasn1)
2) switch haikuporter to master.
3) build libtasn1_x86 on x86_gcc2.
4) reboot to x86.
5) switch haikuporter to the version of this PR.
6) build libtasn1 and put the hpkgs on a safe place
7) rm -rf work-\* for libtasn1
8) rebuild libtasn1 and check that the package obtained previously is identical to thie new package.

But I did not make these tests yet...

waddlesplash:
You don't even need `find`, a simple `rm -rf /path/to/haikuports/*/*/work-*` will do, no?

I don't particularly see the need for such a comprehensive test; it's a safe enough change.

korli:
fbrosson you know you can configure the output directory in haikuports.conf already? That's better than mixing the recipe tree with work directories.
Otherwise the change should be ok.

fbrosson:
@waddlesplash, You are perfectly right!
(How come did I not think of using only rm !!! ;-)

@korli, yes I know, but I like to have the workè\* dirs in the same directory as the recipe, so I don't use that config feature.

I've just updated the commit message, with a notice about the rm command.

Should I merge this PR myself? Or may I leave that to one of you?

korli:
I'm wondering if it wouldn't be better to use a file for the architecture, and to clean up automatically when it doesn't match. You won't ever run Haikuporter for two target architectures at the same time, which is why the secondary arch gets another work directory in the first place.

fbrosson:
I must say I'm not 100% comfortable with the name change in the work-\* directories (because we get longer names), but since it seemed the easiest way, I jumped into it.

@korli, when you say _"use a file for the architecture"_ do you mean a flag file, like the ones we already have? If yes, that was one of my initial suggestions, but I know it can be harder to implement and debug.

> You won't ever run Haikuporter for two target architectures at the same time

True, but we could be using a remote filesystem, and it could be simultaneously shared by several VMs running different architectures.
That could be handy, if e.g. we wish to run several VMs with QEMU and/or VirtualBox, all using the same iSCSI disk to access the shared haikuports tree.

Maybe we should keep this PR open for a few days so that we think about all the pros and cons.
(Needless to say, I'll not merge it without your approval.)

fbrosson:
> Adding more stuff to the work-* dir names will hurt tab completion further, so I lean towards moving them elsewhere or adding an extra directory level. Maybe work-$portVersion/$targetArchitecture$secondaryArchSuffix or something similar?

@pulkomandy, I have to admit that it's a bit annoying to have up 5 work-* directories for each version of a recipe.
So maybe it would be better to follow your suggestion, or maybe this variant:

* **`work-$portVersion/$targetArchitecture`** for primary arch builds,
* **`work-$portVersion/$targetArchitecture-$effectiveTargetArchitecture`** for secondary arch builds,

I'll try this later.

fbrosson:
I see several options:

1. We keep things unchanged and close this PR. Users willing to have separate work- directories will need to define OUTPUT_DIR in their haikuports.conf. (They will probably also want to set PACKAGE_PATH.)

2. We merge only the first commit of this PR. I made a small change so that the work directories will be:
* "work-$portVersion-$targetArch" if building for a primary arch,
* "work-$portVersion-$targetArch-$secondaryArch" otherwise.

3. We merge the two commits of this PR, but the changes are ugly. With this, we have:
* "work-$portVersion/$targetArch" if building for a primary arch,
* "work-$portVersion/$targetArch-$secondaryArch" otherwise.

BTW, with option 3, if I don't do the ugly change in [line 866 of ShellScriptlets.py](
https://github.com/haikuports/haikuporter/blob/39f25c7739c6a85d0ecd10ee7de4aa4bdaf09d81/HaikuPorter/ShellScriptlets.py#L866) then haikuporter fails to unmount the chrooted directory after the hpkgs are built.

4. An easier (but almost as ugly) workaround would be to keep ShellScriptlets.py unchanged, and use:
* "work-$portVersion/work-$targetArch" if building for a primary arch,
* "work-$portVersion/work-$targetArch-$secondaryArch" otherwise.

I'm not confortable with options 3 and 4, and I admit that option 2 is not that good because we end up with  5 work directories for each portVersion.

With option 1, the drawback is that we have to either:
a. clean our work- directories each time we switch from one arch to another,
b. or define OUTPUT_DIRECTORY, but in that case we end up with the work- directories in a different tree as TREE_PATH.

I got used to option 2, but I'm not sure everyone will like it. So we can either keep this PR open until there is some consensus, or close it. (I'm ok with both of these options.)

fbrosson:
After all, let's close this PR and keep haikuporter like it is.

If one want's to not have to worry about removing the work-\* directories everytime one switches to another architecture, then the way to go is to define a unique OUTPUT_DIRECTORY in each haikuports.conf.

Regarding the packages directory, this is even easier. One might set, for example:
* **`PACKAGE_PATH=/path/to/haikuports/packages-x86_gcc2`** on x86_gcc2,
* **`PACKAGE_PATH=/path/to/haikuports/packages-x86`** on x86, and
* **`PACKAGE_PATH=/path/to/haikuports/packages-x86_64`** on x86_64.

So no regrets. Having up to 5 work-\* directories per recipe per portVersion would have been quite annoying.

PR #111, opened 2017-01-19
repository path: Append '-' + buildPlatform.targetArchitecture

fbrosson:
This will make haikuporter use **`repository-<buildArchitecture>`** instead of just **`repository`**, which was being reset and populated again everytime hp detected it was running on another arch.


fbrosson:
This PR addresses half of the first issue discussed in #109.
It works fine for me and can, IMO, be safely be merged.

waddlesplash:
OK by me.

PR #115, opened 2017-02-06
Configuration: new optional REPOSITORY_PATH parameter

fbrosson:
to allow relocation/specification of the repository directory.
~~Of course, this change is neutral if you don't set that parameter.~~

While at it, drop the suffix ('-' + $targetArchitecture) that was added to the default path for the repository directory, in the previous commit. The default path is therefore restored:
$OUTPUT_DIRECTORY/repository

### DIFF
--- a/HaikuPorter/Configuration.py
+++ b/HaikuPorter/Configuration.py
@@ -251,6 +251,8 @@ class Configuration(object):
 			self.outputDirectory = self.treePath
 		if not self.packagesPath:
 			self.packagesPath = os.path.join(self.outputDirectory, 'packages')
+		if not self.repositoryDirectory:
+			self.repositoryDirectory = os.path.join(self.outputDirectory, 'repository')
 
 	@staticmethod
 	def init():
--- a/HaikuPorter/Configuration.py
+++ b/HaikuPorter/Configuration.py
@@ -138,13 +138,13 @@ haikuportsAttributes = {
 		'indexable': False,
 		'setAttribute': 'packager',
 	},
-	'REPOSITORY_DIRECTORY': {
+	'REPOSITORY_PATH': {
 		'type': types.StringType,
 		'required': False,
 		'default': None,
 		'extendable': Extendable.NO,
 		'indexable': False,
-		'setAttribute': 'repositoryDirectory',
+		'setAttribute': 'repositoryPath',
 	},
 	'SECONDARY_CROSS_DEVEL_PACKAGES': {
 		'type': types.ListType,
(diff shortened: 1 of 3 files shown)


fbrosson:
Works like a charm, regardless of whether OUTPUT_DIRECTORY or REPOSITORY_DIRECTORY are defined or not in the configuration.

BTW: if you wish to set OUTPUT_DIRECTORY to something different from TREE_PATH then you will probably want to set DOWNLOAD_IN_PORT_DIRECTORY to "yes", otherwise haikuporter will not look in the download subdirectories it was using when OUTPUT_DIRECTORY was not set.

humdingerb:
@fbrosson , seeing as you're becoming the conf-expert, would you like to add the most important things still missing to the [wiki](https://github.com/haikuports/haikuports/wiki/HaikuPorter-HaikuPortsConf) ? :)

fbrosson:
Good idea. And yes, it looks like most (if not all) of the following parameters should be in the wiki:
- [ ] DOWNLOAD_IN_PORT_DIRECTORY
- [ ] OUTPUT_DIRECTORY
- [ ] PACKAGES_PATH
- [ ] REPOSITORY_DIRECTORY
- [ ] SECONDARY_TARGET_ARCHITECTURES

I'll add them, in alphabetical order, when I have more time. :-)
(I'll check the boxes here when I'm done with them.)

fbrosson:
I made a small change to this PR (see the second commit), and would like to squash and merge it.
May I do so, using the following commit summary?
```
Configuration: new optional REPOSITORY_DIRECTORY parameter

to allow relocation/specification of the repository directory.

While at it, drop the suffix ('-' + $targetArchitecture) that
was added to the default path for the repository directory in
the previous commit. The default path is therefore restored:
$OUTPUT_DIRECTORY/repository
```

fbrosson:
@korli, @pulkomandy, @waddlesplash, and others:
Should I change REPOSITORY_**DIRECTORY** to REPOSITORY_**PATH** and repository**Directory** to repository**Path** ?
I took REPOSITORY_DIRECTORY (and repositoryDirectory) instead of REPOSITORY_PATH (and repositoryPath) for the new parameter because the default value for it is based on that of outputDirectory.
But PACKAGES_PATH, which already exists, is used for packagesPath, and its default value is also based on that of outputDirectory.

So please let me know what you prefer:
1.REPOSITORY_**DIRECTORY** and repository**Directory** (already in this PR)
or
2.REPOSITORY_**PATH** and repository**Path** (the change is easy to make, and PATH is shorter than DIRECTORY ;-)

Please also let me know if that's acceptable to drop the suffix I had recently added in #111 ... (Well, I guess it is.)
Hmm, I need to say that if we want to keep the suffix then we cannot keep the second commit of this PR, because Configuration.py does not know what the targetArchitecture is, if that's not in the config.

Thanks!

PR #127, opened 2017-06-21
fix for --update-dependencies, support for --list --print-filenames.

korli:


diversys:
Would be nice to have `--list --print-filenames` by default

Issue #121, opened 2017-03-10
Possible conflict of packaged mime_db


humdingerb:
haikuporter extracts the supported types from the ressources of the to-be-packaged binary and adds the respective supported types entries as $dataDir/mime_db/*.
It seems to only add the MIME types that aren't already listed in the system's mime_db. Take ArtPaint, it's .rdef declares:
```
	"types" = "image/gif",
	"types" = "image/x-iff",
	"types" = "image/jpeg",
	"types" = "image/png",
	"types" = "image/x-targa",
	"types" = "image/x-iff-ilbm",
	"types" = "image",
	"types" = "image/x-tiff",
	"types" = "application/x-vnd.artpaint-project-file",
        "application/x-vnd.artpaint"
```
In the package, there's only:
```
	"types" = "image/x-tiff",
	"types" = "image/x-iff-ilbm",
	"types" = "image/x-iff",
	"types" = "application/x-vnd.artpaint-project-file",
        "application/x-vnd.artpaint"
```
That's probably alright.
What isn't, is that the packaged  $dataDir/mime_db/image folder includes no attributes for supertype "image", no icon and extra attributes like Width, Height, or preferred app. Installing ArtPaint will "overwrite" the system's mime_db/image resulting in all image files missing their icon and extra attributes.

Becasso, which also declares supported types
```
	"types" = "image/x-becasso",
	"types" = "image",
	"types" = "application/x-sum-becasso.palette"
```
, does not show the same issue of removing the image supertype's attributes...

I hope I have described the issue adequately. ArtPaint and Becasso are available in the HaikuDepot.
(I have also tried building ArtPaint after removing the "types" = "image" from the rdef file, but that had no effect, as the "image" folder will be created in any case.)

korli:
Should be workarounded/fixed with 76cd2c9a871d452ef64fc9708621cb081ed6a088

PR #130, opened 2017-07-21
Policy: support for relative RPATH, Package: touch supertypes in the past

korli:


korli:
This PR proposes a fix for #121.

Issue #87, opened 2016-03-25
Print path when extracting patchset


diversys:
hp -e htop_x86 prints
Extracting patchset for 2.0.1.tar.gz
Would be handy to have a full path:
Extracting patchset for 2.0.1.tar.gz to /boot/home/haikuports/sys-process/htop/patches/htop_x86-2.0.1.patchset

PR #149, opened 2017-11-29
Print path when extracting patchset #87

DrRamm:
Or I didn't get you?

https://github.com/haikuports/haikuporter/issues/87


DrRamm:
print ('Extracting patchset for %s to %s' % (self.fetchTargetName, patchSetFilePath))

this should be more clearly

PR #157, opened 2017-12-22
HaikuPorter: use cpu_count() as default value for -j

owenca:
Also, change -j default value from 1 to the CPU count.

Fixes #1039

### DIFF
--- a/HaikuPorter/Options.py
+++ b/HaikuPorter/Options.py
@@ -17,6 +17,7 @@ from .__version__ import __version__
 from .Utils import isCommandAvailable, warn
 
 from optparse import OptionParser
+from multiprocessing import cpu_count
 
 
 # -- global options -----------------------------------------------------------
@@ -77,7 +78,7 @@ def parseOptions():
 					  action='store_true', dest='yes', default=False,
 					  help="answer yes to all questions")
 	parser.add_option('-j', '--jobs',
-					  action='store', type="int", dest='jobs', default=1,
+					  action='store', type="int", dest='jobs', default=cpu_count(),
 					  help="the number of concurrent jobs to build with")
 	parser.add_option('-S', '--strict-policy',
 					  action='store_true', dest='strictPolicy', default=False,


fbrosson:
You might wish to split your PR into 2 parts. One for the change regarding the number of CPUs, and another for enabling the strict-policy mode. (I'd move the second change to a new PR.)

It's a very good idea to use the number of CPUs as default value for `-j` and I vote for it :)

However, for the second change, if you enable strict-policy by default but do not provide an option to disable it, then you are just removing a functionality, i.e. the ability to run haikuporter in _normal_ mode.
I always use strict policy because this helps finding missing/unwanted lines in PROVIDES and REQUIRES, but I do this with a shell alias.
But there can be situations where one might wish to have the hpkg created before all policy-related issues are fixed.
We could change:
```
  -S, --strict-policy   require strict packaging policy adherence; packaging
                        will fail on any policy violation
```
to
```
  -S, --no-strict-policy   do not require strict packaging policy adherence; packaging
                        will not fail on policy violations
```
or use different switch options if anyone has better suggestions.
What do you think? Any other opinions?

pulkomandy:
Even worse, there are packages which can't be built in strict mode. This can happen because of limitations in haikuporter, for example if libraries are stored in a non-standard path in the package but binaries are linked to them. This seems to be the case for openjdk, and I think it used to be for llvm as well.

--get-dependencies should not be enabled by default, because it changes your running system. We will make it the default only after it has been fixed to put the dependencies in haikuports/packages.

fbrosson:
Regarding the change about the `--get-dependencies` option, I have the same opinion as for the `--strict-policy` option.
(I therefore totally agree with @pulkomandy.)

owenca:
Now I see the problem although [Issue 1039](https://github.com/haikuports/haikuports/issues/1039) asks for the changes. Please see the updated commit [
1a4266a](https://github.com/haikuports/haikuporter/pull/157/commits/1a4266a1d6a0a0b6751d59c41b19b6c18bb346f4).

fbrosson:
I'm currently running Haiku-x86_gcc2 on a Xen domU with 2 VCPUs.
I did notice that the patched haikuporter correctly finds the number of CPUs. Indeed, I see a "make -j2" when I call haikuporter without the `-j`. And I see "make -j1" if I pass a `-j 1`.
This is why I think this PR is OK.

I have, however, added a `wip-no-merge` label for 2 reasons:
* I'm having some trouble building with anything else than `-j 1` and I don't know yet if this is due to my VM or to the packages I am trying to build, or if this is an unexpected downside of the patch.
* I don't know if the build servers are already using `-j x` with x>1. If this is the case then that's OK. But if they are not passing any `-j` then they will start using more than one CPU if they use the updated haikuporter. This might trigger build failures for those packages which might not have been detected as incompatible with `$jobArgs`.


@owenca, I noticed that you have based this PR on your master branch. Although I have pulled your remote master "branch" (into a local "cpu" branch on my side and would therefore be able to push it to the projects master branch), **if** you delete your haikuporter repository **then** we won't be able to merge this PR. It would have to be closed and we would either need to open another PR or push your commit from my local copy of your commit. So please do **not** delete your repo before we have merged this PR. Thanks!

owenca:
@fbrosson, I won't touch https://github.com/owenca/haikuporter until you are done. I will try to use a feature branch in the future when I feel confident.

fbrosson:
@owenca, there seems to be a bug in Haiku or in my VMs. The build freezes everytime I try to run `haikuporter-j n`... with n\>1.
I'm having this problem with both QEMU+Xen and QEMU+KVM. I haven't tried with VirtualBox yet.

So if we merge this PR as is, other users might have the same freezes I'm experiencing.
(The system does not freeze, it's just haikuporter who freezes.)
Your patch is **not** to blame, because I still have the freeze without the patch. Of course, it only happens if I pass `-j 2` (or bigger), no matter how many CPUs or virtual CPUs I have.

Hence this question: would you like to try and patch Configuration.py (and maybe a few other files too) to teach haikuporter it has to read a new setting, say `AUTODETECT_CPU_COUNT`, to decide which default value it should use for the number of jobs? It would choose 1 if that configuration parameter is set to false (or absent), or `cpu_count()` if that config is set to True.

You would have to insert this in Configuration.py, after ALLOW_UNSAFE_SOURCES:
```
        'AUTODETECT_CPU_COUNT': {
                'type': YesNo,
                'required': False,
                'default': False,
                'extendable': Extendable.NO,
                'indexable': False,
                'setAttribute': 'autodetectCpuCount',
```
You would also need to patch the __init__ function in that same python script:
```diff
 		self.allowUnsafeSources = False
+		self.autodetectCpuCount = False
 		self.downloadInPortDirectory = False
```
later:
```diff
 	@staticmethod
 	def shallAllowUnsafeSources():
 		return Configuration.configuration.allowUnsafeSources
 
+	@staticmethod
+	def shallAutodetectCpuCount():
+		return Configuration.configuration.utodetectCpuCount
+
 	@staticmethod
 	def shallDownloadInPortDirectory():
 		return Configuration.configuration.downloadInPortDirectory
```
Then you would have to make a few extra changes to add the required glue between Configureation.py and Options.py.
I made similar changes in 180b2367fbad451b6eb08101098c970577c920ad last february and it was for a very small change, so it would be easy to figure out where you need to add the glue code.
Of course, if you take this challenge, your GCI task will count double ;-)

Even if the freezes occur on all Haiku systems, this improved patch is still a nice feature because haikuporter is also usable on Linux.

To all: needless to say, any suggestions regarding the names (or anything else) are welcome!

fbrosson:
Has anyone recently built a package on Haiku using `haikuports -j 2` (or greater) ?
(No need to pull the patch in this PR for that test.)
Does the build succeed?

I know I already used `-j 2` one year ago, without any problems, but I'd like to know if I'm the only one to experience the freezes I described above. Thanks!

If yes, then we don't need to bother @owenca with the additional changes I suggested above.
And we can also merge this PR if everyone agrees.

pulkomandy:
I don't have problems with -j4. You should open a bugreport to haiku with that problem.

fbrosson:
Thanks @pulkomandy for the info. Does it work for you on all archies?

I tested on x86_64.and got no freezes. And the patch in this PR is working fine for me (on that arch.)

So I guess my x86_gcc2 is broken.
If someone could try building something on x86_gcc2 with `haikuporter -j 2` (or bigger) and let me know whether or not it freezes, that would be great! Thanls!

owenca:
@fbrosson, I have implemented most of the changes for --autodetect-cpu-count in case it's needed.

owenca:
I don't have a problem using -j2 on x86_gcc2 in VMWare Fusion.

mmlr:
All 4 builders at kitchen.haiku-os.org are set up to use `-j2`. They would consequently not be affected by this change. Considering the thousands of packages they already built, they don't seem to be affected by any freezing with parallel builds and neither do most recipes seem to be affected by undetected parallel build issues.

fbrosson:
@owenca, this is great! I hope we will take your extra changes for `--autodetect-cpu-count` !

Thanks @mmlr and @pulkomandy for the info. I really think we should also include the extra changes Owen has almost finished. I was planning to merge this PR and then implement the extra changes myself and open a PR for them, but since Owen is offering them if needed, it would be sad not to have them.

This option will be very handy in several scenarios:
* to use the maximum number of processors without even having to hardcode a "-j4" (or anything else) in the configuration.
* for users who, for some unknow reason, would need to pass `-j 1` to avoid the freezes.

Shall we invite Owen to push his extra changes to a separate branch on his remote repo so that we can see if we it would be OK to include it in this PR?

@owenca, I hope you did not spend too much time on it. Even if you did not finish it yet, it would be interesting to see how you are implementing the extra option. If you have almost finished the extra option (or wish to do so) then I'll try to convince the other controbutors that we should include your extra changes.

Thanks you all!

fbrosson:
Hmm, we can merge this PR right now, after all.
There are now downsides (other than me having to use `-j 1` on x86_gcc2 after the merge).

@owenca, don't worry about the extra patch. I hope you did not spend too much time on it.
BTW, I think you were preparing something more complete than what I had in mind.
I was only thinking about adding support for a config parameter in the haikuports.conf file.

owenca:
@fbrosson, I will put it on hold then.

fbrosson:
Yes please, that would be cool.
I have my idea on how to add a AUTODETECT_CPU_COUNT parameter in `~/config/settings/haikuports.conf`.
When you wrote:
> I have implemented most of the changes for --autodetect-cpu-count in case it's needed.

do you mean that you added a commad line option for haikuporter?
This is not what I have in mind. It is better to not add such an option, but, instead, add a config parameter.

This is what I have in mind (and in my "cpu" branch, now ;-)
https://github.com/haikuports/haikuporter/compare/master...fbrosson:cpu

But it's not finished. I would need to change Options.py to call `Configuration.shallAutodetectCpuCount()` to decide which default value should be used for the `-j` parameter, that being `1` for False, or `cp_count()` for True.

owenca:
@fbrosson, I thought you wanted it in both the command line and the config file. It would be much simpler if we have to deal with the config file only.

fbrosson:
After all, you are perfectly right. You have convinced me.
It can be handy to have a command line option. Then anyone would be able to create an alias, say, hp2, which would be the same as the hp alias, but with the `--autodetect-cpu-count` option.

Shall we merge my small change with what you have already prepared?
(You can copy-paste my changes into a single commit on a new PR.)
Hmm, don't forget to change the default value of the AUTODETECT_CPU_COUNT to false, otherwise the command line option would not be usefull.

**But before you work on this any further**, it would be better to wait for comments from the other key contributors. I hope they will buy the idea of the config and command line switch, but I can't decide for them :-)

Comments welcome!

mmlr:
I don't really see the benefit of adding a configuration or option for this feature. It's not exactly hard to figure out the CPU count, so writing `-jX` is a lot more compact than remembering and typing `--autodetect-cpu-count`. Also if I have to add something to the configuration to make this work it's rather pointless, because it'd again be simpler to add the corresponding `-j` to an alias I'm already defining anyway.

If a configuration gets added I'd much rather have `JOB_COUNT` which would provide the default value for `-j` and could have a possible value of (and default to) `auto` which would cause auto detection.

fbrosson:
Eureka!
How about a JOB_COUNT (or MAX_CPU or NUM_CPU or CPU_COUNT or anything more appropriate) in haikuports.conf which would be optional, have 1 as default value and which we could set to:
* 0 to pass cpu_count() to `-j`
* or any other number for `-j`

On the command line, we could say that if haikuporter is called with -j 0 then it has to call cpu_count() and use that for `-j`.

The advantages of the config paramter is that if the system is on a removable device, you don't need to edit the config if you use your removable device on different machines (with different number of CPUs) and automatically use the maximum number of CPUs.

The advantages of the `-j 0` switch is that you can put that in a shell alias and use the same alias on all your systems.

Both of these methods (config and command line option) are very handy if you change the number of virtual CPUs in your VM. Everything is automatically adjusted!

This also brings flexibility because you can, for instance, keep the same shell alias on all your "flavours" of Haiku, by using a system wide default JOB_COUNT.

The idea of setting `JOB_COUNT=0` to use all the CPUs is great, isn't it?

What do you think? If that's OK, what name would you like to have for the config parameter?

fbrosson:
Sorry @mmnl, I just realized that what I described above is almost what you wrote before, except that I'm suggesting 0 (zero) instead of "auto". So my _"Eureka!_" was inappropriate :)
Actually, when the idea stroke my mind I had not finished readling your message!

So, given that what I'm suggesting now is almost what you described, would you be OK with this:

* JOB_COUNT in haikuports.conf with 1 as _default_ value, and 0 as _magic_ value to trigger the use of all CPUs.
* handling of `-j 0` (instead of the `-jX` you suggested).

Zero as limit is easier to remember because it is commonly used as synonym of _"unlimited"_.

Can we give @owenca green lights so he can make changes to the work he has started, if he wishes?

Thanks!

Issue #151, opened 2017-12-06
support for TEST_REQUIRES ?


fbrosson:
When a recipe has a `TEST` function which needs commands that are not needed during `BUILD` we still have to add them to `BUILD_PREREQUIRES`.
It would be nice to be able to take out of `BUILD_PREREQUIRES` all the commands which are not required for `BUILD` and "move" them to a separate variable, say, TEST_REQUIRES or TEST_PREREQUIRES.

Suppose that we need md5sum and sha256sum for running `TEST`, and that we don't need them in `BUILD`. We would add this after `BUILD_PREREQUIRES`:
```
TEST_REQUIRES="
	cmd:md5sum
	cmd:sha256sum
	"
```
Then, we would (currenly need to) add this hack:
```
if [ "$recipeAction" = TEST ]; then
	BUILD_PREREQUIRES="$BUILD_PREREQUIRES
		$TEST_REQUIRES
		"
fi
```
It would be cool to teach haikuporter that when it is running the tests, it should examine `BUILD_PREREQUIRES` and `TEST_REQUIRES`. This would, of course, allow us to remove the hack above.
(This hack does not appear like this in any recipe, although app-arch/tar uses a `if [ "$recipeAction" = TEST ]; then` to append lines to BUILD_PREREQUIRES.)

I read _listBuildDependencies in Main.py, and:
* resolveBuildDependencies
* _doBuildStage
* _doTestStage

in Port.py but did not figure out how to modify haikuporter to teach it how to handle a TEST_REQUIRES.

BTW, I don't know what name would be most appropriate, between `TEST_REQUIRES` and `TEST_PREREQUIRES`.

Since haikuporter has received several great improvements recently by Python experts, I decided to file this issue.
Thanks!

alaviss:
We really need this. A port I'm making (atf) requires (kyua) for it's `TEST()`, but the latter also requires the former to build

fbrosson:
When I wrote the description of this issue I though that we would need to choose only one variable name, and I was suggesting to choose between `TEST_REQUIRES` and `TEST_PREREQUIRES`.
Well, it turns out that we need both of them.

I also thought we could workaround the chicken/egg issue by using a
```
if [ "$recipeAction" = TEST ]; then
	BUILD_REQUIRES="$BUILD_REQUIRES
		$TEST_REQUIRES
		"
	BUILD_PREREQUIRES="$BUILD_PREREQUIRES
		$TEST_PREREQUIRES
		"
fi
[...]

PR #160, opened 2018-04-10
Do not make the TEST phase depend on REQUIRES.

fbrosson:
Before the introduction of the `TEST_REQUIRES` variable, the `TEST` phase depended only on `BUILD_REQUIRES`. After `TEST_REQUIRES` was introduced, the `TEST` phase started to _also_ depend on `REQUIRES` and `TEST_REQUIRES`. The downside of making the `TEST` phase depend on `REQUIRES` is that, in some cases, haikuporter cannot satisfy such a dependency. That's because haikuporter considers the `REQUIRES` of the base package and _every_ subpackage. This means that if any subpackage has a dependency that cannot be met then `--test` becomes useless for that package.

The fix for this issue is easy:  make the `TEST` phase depend _only_ on `BUILD_REQUIRES` and `TEST_REQUIRES`.

This change does not break anything. It only means that, from now on, we will have to consider adding `TEST_REQUIRES="$REQUIRES"` if the `TEST` phase needs it. Hopefully, it seems we currently don't have any recipe that would need this.

Another advantage of this change is that the packages needed to run `haikuporter --test [--enter-chroot] <some-package>` will be restored to what that list was before the introduction of the `TEST_REQUIRES` variable, with the exception, of course, of what is mentioned in that variable.

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -1328,7 +1328,7 @@ class Port(object):
 		   environment (chroot) for the test stage"""
 
 		return self._getNeededPackages(packagesPath,
-			['REQUIRES', 'BUILD_REQUIRES', 'TEST_REQUIRES'],
+			['BUILD_REQUIRES', 'TEST_REQUIRES'],
 			'required packages for test', True)
 
 	def _executeBuild(self, makePackages):


fbrosson:
@mmlr, shall I merge this?
It addresses the main issue spotted in [this comment](https://github.com/haikuports/haikuporter/issues/151#issuecomment-378401942)
Thanks!

PR #190, opened 2019-09-05
Add support for ARM64 packages

jarekpelczar:
If architecture is named aarch64 this is remapped to arm64.

Signed-off-by: Jaroslaw Pelczar <jarek@jpelczar.com>


waddlesplash:
I don't think the remapping is necessary here, internally this should all be ARM64.

jarekpelczar:
I removed the "aarch64" mapping

waddlesplash:
Appears this was already done in ddfc743e1acaf4dfefc5ad5bbc41352fb6df3eef. So this PR is redundant.

PR #200, opened 2020-02-02
python3: errors from mutating collections while iterating

### LINKED ISSUE #199: dictionary changed size during iteration
After pulling haikuporter tree and haikuports tree, haikuporter gives me this:
```
~> haikuporter
dictionary changed size during iteration
```
**EDIT** this is on the gcc2h (did a pkgman full-sync before, rebooted, pulled (cleaned leftover source and downloads)) and then ran haikuporter

ambroff:
This function has two loops which iterate over the items in a dictionary and
then conditionally remove from that dictionary. dict.items() returns an
iterator and not a copy of the (key, value) pairs it contains, so it isn't
safe to remove from this loop.

This makes a copy of the items by converting the iterator to a tuple.

Fixes #199.

### DIFF
--- a/HaikuPorter/Repository.py
+++ b/HaikuPorter/Repository.py
@@ -298,7 +298,7 @@ class Repository(object):
 		# are supported, but we won't know until we have parsed the recipe file.
 		secondaryArchitectures = Configuration.getSecondaryTargetArchitectures()
 		if secondaryArchitectures:
-			for port in self._allPorts.values():
+			for port in tuple(self._allPorts.values()):
 				for architecture in secondaryArchitectures:
 					newPort = Port(port.baseName, port.version, port.category,
 						port.baseDir, port.outputDir, self.shellVariables,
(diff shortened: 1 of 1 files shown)


Begasus:
Confirmed working again, thanks!

Begasus:
Could you add a fix for line 301 to it? (that fixes it for gcc2h)

ambroff:
Ah yeah I figured more would be broken. I only fixed the part that crashed for me. Let me audit the rest to see if there are others aside from the one you pointed out @Begasus.

Begasus:
No problem, commented on the issue with --debug enabled, thanks for looking into this!

ambroff:
I read through the repository module and I don't see any other similar errors. I spent some time skimming through the rest of the code and it looks good I think.

kallisti5:
Thanks for the work here @ambroff !!!

PR #202, opened 2020-02-13
WIP support for query-based menus in Deskbar

mmuman:
I had this around for years… Today's discussion on IRC made me think I should start the ball rolling on how to scale the apps menu up.

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -595,6 +595,11 @@ addAppDeskbarSymlink()
 		entryName=$1
 	fi
 
+	# add attributes for query-based Deskbar menus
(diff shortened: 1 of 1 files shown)


mmuman:
I already had [a try in ZETA](http://revolf.free.fr/beos/shots/shot_query_based_menu.png) back then, no reason it wouldn't work in Haiku…

kallisti5:
interesting. @mmuman does a "deskbar" side of this exist somewhere?
My deskbar menu is getting unruly on my desktop lately.  Seems like quick access menus could also leverage this.

mmuman:
The Deskbar side only consists of an Application folder with queries on the categories and some more checks…

mmuman:
It might have some security implications though, so we might want to have some restrictions on which filesystem to query from at some point…

PR #209, opened 2020-08-10
Policy: check if a package provides itself

mmuman:
This is a common mistake, like when the package name is different than the provided command, it just happened to me again on hteditor.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -104,6 +104,11 @@ class Policy(object):
 		return names
 
 	def _checkProvides(self):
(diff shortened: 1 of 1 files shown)


alaviss:
How about we just add this by default to all packages? Is there any case where this is not needed?

mmuman:
I don't know.

pulkomandy:
No, a package without this provide will only create problems. +1 to adding by default.

waddlesplash:
Packages should declare their own provides with potential `compat` versions; we cannot guess that directly. So an error is most correct.

PR #214, opened 2021-02-24
HaikuPorter/BuildMaster: MongoDB reporter

kallisti5:


kallisti5:
Trying to make the data in haikuporter buildmaster more reasonable to be parsed with a web app.  We're still grepping 1000's of json files to find last package builds, etc.

builders:
builders

buildruns:
buildruns

kallisti5:
The code is pretty simplistic and tries to not arbirararly modify dictionary layouts.

Benefit is actually being able to rapidly search 1000's of buildruns for packages via an index of the json structure.  Obviously more needs done here to render a web application based on the mongodb contents.

A json "reporter" allows continued "dumping" buildmaster state to on-disk json files.

kallisti5:
Updated this one for the latest code. Fixed some old calling conventions to pymongo.

The data format sucks since it's still the internal format used by haikuporter, but we could "translate" it into something easier to consume within the "ReporterMongo" code.   We could also translate the data into something easier to consume within the ReporterJson code.

For now though, this one opens the door to reporting to things other than 10,000 json files on disk which are slow to parse.

buildruns
builders

waddlesplash:
Do we have a way to import old JSONs into Mongo, too? It would be good to be able to search all old builds...

kallisti5:
It makes some arbitrary adjustments to the dict.. so we would need a short tool to translate the existing json to the intended mongo layout.

kallisti5:
Based on my conversations with waddlesplash, this represents a big improvement in options available for making a web interface for buildmaster.

I'm going to go ahead and merge it.

waddlesplash:
TBH the reporter files should go in a subdirectory.

I'll likely take a look at SQL-izing this at a later date and refactoring how status info is reported. Doesn't seem like it would be too hard...

PR #222, opened 2021-11-17
BuildPlatform: assume 'which' command is available on Unix build host

davidkaroly:
Latest change in haikuporter introduces cmd:which as a prerequirement for shell scriptlets.
When I try to do bootstrap build, now I get the following error message:

`scriptlet-prerequires "cmd:which" of package "gcc_bootstrap-8.3.0_2019_05_24" could not be resolved`

This proposed change adds cmd:which to implicitBuildHostProvides when cross-building.


waddlesplash:
Thanks!

mmuman:
FTR on Debian Sid `which` complains it's getting deprecated:
```
/usr/bin/which: this version of `which' is deprecated; use `command -v' in scripts instead.
```

PR #223, opened 2021-12-14
Cleanup builder logic

kallisti5:
This is some initial cleanup of the builder logic before attempting automatic spin-up and down of workers at Digital Ocean


kallisti5:
tested working locally connecting to a single Haiku x86_64 vm and running a build

HaikuPorter/BuildMaster.py
```diff
@@ -843,14 +212,15 @@ def __init__(self, portsTreePath, packagesPath, options):
 
 				builder = None
 				try:
-					builder = RemoteBuilder(configFilePath, packagesPath,
```
kallisti5, line 846: eventually, i'm going to set a "builder type" field to the builder config and choose from the supported builder types (local, ssh, digitalocean, vultr, etc)

HaikuPorter/BuildMaster.py
```diff
@@ -813,12 +188,6 @@ def __init__(self, portsTreePath, packagesPath, options):
 
 		self.localBuilders = getOption('localBuilders')
 		self.remoteAvailable = False
-		if paramiko:
-			self.remoteAvailable = True
-		else:
-			print('Remote mode unavailable')
```
kallisti5, line 819: I dropped this logic because it seemed unnecessary.  Local Builders appear to be for running buildmaster natively on Haiku building locally.

localBuilder here is *not* running ./haikuporter on haiku.  This is purely --build-master logic.

The remoteAvailable = true moved down to after successfully adding a remote builder.

kallisti5:
I've successfully built several packages in remote ssh mode with this change as it stands.  I think the final bugs have been worked out.

pulkomandy:
Looks ok to me, maybe @korli wants to review? IIRC he was using the client/server mode also outside of buildmaster?

korli:
Looks good to me

PR #226, opened 2022-01-08
increase maximum allowed length of SUMMARY to 100

jmairboeck:
this was discussed in haikuports/haikuports#6353

it came up here: https://github.com/haikuports/haikuports/pull/6229#discussion_r780248489

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -390,8 +390,8 @@ class Port(object):
 		if entries[key].count(' ') < 2:
 			sysExit('%s must have at least 3 words (%s).'
 					% (key, self.recipeFilePath))
-		if len(entries[key]) > 80:
-			sysExit('%s exceeds 80 chars (in %s)'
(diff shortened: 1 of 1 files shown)


Begasus:
As mentioned in the comments there, I'm not sure that for SUMMARY we should extend this to 100 characters (just my 2 cents) :)

jmairboeck:
If this should not be changed, we can also close this. This is just a suggestion.

Begasus:
Let's wait what others have to say about it, it's good to bring this up :+1:

humdingerb:
I agree to keep the limit of the SUMMARY. Its intention is to keep the summary short and succinct. That also ensures that the summary column in HaikuDepot can be kept nice and narrow.

Begasus:
I mentioned this too in the comments, in FileTypes I think it would look a bit overkill (when used in rdefs) :)

humdingerb:
Alright, let's keep the SUMMARY max length as is and close this.

PR #232, opened 2022-09-14
Add generic python recipes

Begasus:
I've been using this generic recipe for a while now, could be handy for others too?


waddlesplash:
IMHO, we should do something about the "duplication" of all this logic in every Python recipe.

Perhaps we should have a "templates" directory in the HaikuPorts repository with scriptlets/functions for specific recipe types, which could then be invoked and take care of the general logic here?

CC @korli, @oscarL

Begasus:
Made some changes to the default python recipe and added one for those using whl archives and one for using build/installer in the source.
For me they are a nice thing to have around to create a recipe for python packages quickly.

PR #239, opened 2023-02-18
use relative paths in cmakeDirArgs

davidkaroly:
Seems to work better than using absolute paths everywhere.
I tried it with wireshark and tenacity, I get successful builds without any patching on the `CMakeLists.txt` files.

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -1212,8 +1212,11 @@ class Port(object):
 				for k, v in configureDirs.items())
 
 		# add one more variable containing all the dir args for CMake:
-		cmakeDirArgs = {}
-		for k, v in configureDirs.items():
+		cmakeDirArgs = {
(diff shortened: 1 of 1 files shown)


HaikuPorter/Port.py
```diff
@@ -1212,9 +1212,23 @@ def _updateShellVariables(self, forParsing):
 				for k, v in configureDirs.items())
 
 		# add one more variable containing all the dir args for CMake:
-		cmakeDirArgs = {}
-		for k, v in configureDirs.items():
```
pulkomandy, line 1216: looks like a simpler fix would be replacing configureDirs with relativeConfigureDirs? It would avoid duplicating all the paths here.

davidkaroly: it's not exactly relativeConfigureDirs because we also have prefix and sysconfdir which are absolute

PR #243, opened 2023-03-07
Introduce runCmake

davidkaroly:
Similarly to configure, sometimes we don't want to pass all the `CMAKE_INSTALL_xDIRS` to CMake. Adding `runCmake` that allows doing this using argument `--omit-dirs`

e.g. for wireshark:
```
runCmake --omit-dirs CMAKE_INSTALL_DATADIR \
    -S . -B build -G Ninja \
    -DCMAKE_EXE_LINKER_FLAGS="-lnetwork" \
    -DCMAKE_BUILD_TYPE=Release
```

This way I can build wireshark with correct dirs (it has its own non-standard default value for DATADIR what would have been overridden by `cmakeDirArgs`)

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -1213,18 +1213,23 @@ class Port(object):
 
 		# add one more variable containing all the dir args for CMake:
 		cmakeDirArgs = {
-			'PREFIX': prefix,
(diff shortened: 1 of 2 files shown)


waddlesplash:
I'm not sure this is a good idea, given how non-standard CMake setups are compared to the regularity of autotools. @Begasus' suggestion may be better.

PR #244, opened 2023-03-07
make unreferenced patches an error

Habbie:
so that the PR lint workflow will go red if this happens


Begasus:
This brakes recipes on 32bit where a patch is only applied for primary architecture, one example is librsync:

```
Checking if any dependency-infos need to be updated ...
        updating dependency infos of librsync-2.0.0
        recipe for librsync_x86-2.0.0 is still broken:
        Error: Patch file librsync-2.0.0.patchset is not referenced in PATCHES, so it will not be used
```

PR #247, opened 2023-03-22
Revert "HaikuPorter ShellScriptlets: Validate there is only either a …

kallisti5:
…shared or a static library."

* This is expected.  Packages can definitely ship static and shared libraries.
* Static libraries get placed into $developLibDir via prepareInstalledDevelLib for example:
  * libcurl.a gets moved to develop/libs
  * libcurl.so* gets left in libs
* gcc / ld chooses a static library via -static

This reverts commit 8a5746030dcf520390af89b46cd0a6aea6c72499.

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -428,15 +428,6 @@ prepareInstalledDevelLib()
 		fi
 	done
 
-	# Make sure there is not a static library in addition to a shared library.
-	if [ -f "$sharedLib" ] \
-			&& [ -f "$installDestDir$libDir/$libBaseName.a" \
-				-o -f "$installDestDir$developLibDir/$libBaseName.a" ]; then
-		echo "prepareInstalledDevelLib error:" \
-			"there is both a shared and a static library for $libBaseName!"
-		exit 1
-	fi
-
 	# Move things/create symlinks: The shared library file and the symlink for
 	# the soname remain where they are, but we create respective symlinks in the
 	# development directory. Everything else is moved there.


waddlesplash:
No, we *don't* want to do that. If curl (a good example) ships both libraries, how will we ever know what dependent packages must be rebuilt when curl has a CVE, for instance, or has to be patched for some other reason? If we enforce shipping only one or the other, then it is VERY clear whether dependent packages must be rebuilt. For packages like `curl` especially, we really want there to be only the shared copy, so updating it does not require mass-rebuilds (or worse, missing rebuilds because 1-5% of packages which need `curl` are static-linked while the rest are dynamic-linked.)

We have had policies to discourage static linking for a long time. This change merely elevated it to the level of an enforced policy. We have been doing this for multiple months, you will note the change in question was enacted before beta4, and things have been working fine. In fact I think there were somethings which updated or went more smoothly because we could now actually rebuild one package instead of many.

kallisti5:
Everyone had a long chat in irc about this.

Here's the general consensus:
* Static libraries are bad. I'm and old man for thinking otherwise
* Static libraries create an unknown upgrade path for any ports that choose them (even mistakenly)
* Any binary-application developers (aka game or proprietary software devs) should be vendoring libraries and not relying on our os-provided static libraries.
* -static is quietly non-functional on Haiku for quite some time in the gcc spec, so leveraging them is difficult. (waddlesplash mentioned -l:coollib, but I didn't confirm it because i'm lazy)

Given the above, i'm quietly abandoning this one.   I guess static libraries are dead to us :cry: :tulip: :skull: :zombie: :gun: :skull:

kallisti5:
@waddlesplash though.  Here's one final thought.   What if something expects and needs to ship static libraries as well as shared ones?   (gcc? clang? llvm?)  there maybe a need for an override somewhere.

waddlesplash:
There are mechanisms to override already. And there are packages that ship static libraries as well as shared ones, like gcc, indeed; but none of the static libraries have the same name as shared libraries.

PR #250, opened 2023-03-25
ShellScriptlets: Fix meson setup command

### LINKED ISSUE #249: 'meson setup <builddir>' fails in recipes
Recipe: https://github.com/haikuports/haikuports/blob/073bae46af6d69fc51db735aa457f1650f7377dc/media-libs/gst_plugins_good/gst_plugins_good-1.22.0.recipe
Log: https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/8335/builds/79082.log

Although the recipe used the recommended `meson setup build` command, rather than the deprecated `meson build`, the porter failed. The `--wrap-mode=nodownload` argument inserted [here](https://github.com/robxnano/haikuporter/blob/adc25c2375d3ad59b4f1f6874abd0cd56d8a6538/HaikuPorter/ShellScriptlets.py#L354) caused it to interpret `setup` as the name of a directory.

This needs to be fixed since the old syntax will be [removed](https://github.com/mesonbuild/meson/commit/3c7ab542c0c4770241eae149b0d4cd8de329aee0) at some point in the future.

korli:
Put --wrap-mode=nodownload at the end of the arguments list, to avoid conflicting with the meson setup command. To allow recipes to override, only use it when wrap-mode isn't in the arguments. Fixes #249.

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -351,7 +351,11 @@ meson()
 		MESON=$(which meson)
 	fi
 
-	$MESON --wrap-mode=nodownload "$@"
+	if [[ "$1" == setup && "$*" != *wrap-mode* ]]; then
+		$MESON "$@" --wrap-mode=nodownload
+	else
+		$MESON "$@"
+	fi
 }
 
 fixDevelopLibDirReferences()


robxnano:
Sorry, but this isn't ready yet as it doesn't take into account existing recipes that don't use the setup command. I haven't found a way to make it work for both commands yet.

korli:
Superceded by #250

PR #254, opened 2023-04-30
Support commented out lines in the PATCHES section.

OscarL:
We could have commented out lines in PROVIDES, and REQUIRES, but not in PATCHES, as those lines weren't being filtered out, ending on "file not found" errors further down.


Begasus:
A little background on the merge from my side, I tend to grab patches (or search them) from other resources, check them out and sometimes they don't apply, I always had to remove the mentioned line to continue.
This is especialy handy for these things, not specific to buildmasters. Thanks @OscarL !

PR #259, opened 2023-07-11
Create modern python package for HaikuPorter

jurgenwigg:
Main goal is to create proper python package which could be installed via `pip install` command. At this moment package is created in a such way that it is not possible to install package flawless on newer python versions.

Creating unit tests, doing cleanup and removing all python2.7 leftovers, enabling lint, types and other static code analysis will be done under separate PR.


scottmc (dismissed):
Why the rename?  I really hate having_under_scores_in_files_names
It's been called haikuporter since the day it was created.

jurgenwigg:
@scottmc I had to rename file due to issue I had on my computer. MacOS is case insensitive and it caused that while cloning the repository I had a big problem. File names will be as in the original repository - haikuporter. But file itself will be **.py** file instead of **HaikuPorter** or I can create directory for the source code of haikuporter itself. Both options are good but it depends on our decision.

In terms of future changes, for Python PEP8 suggests pretty clearly how to style names for classes, functions, etc. Enabling checking formatting for the python code would require setting custom RC file for pylint or changing the names to comply with PEP8 standard. This is open point for future PR.

pulkomandy:
with some reorganization of the sources it should be possible to keep the name as "haikuporter", lowercase, for both the file and the directory. The file is meant to be installed in some /bin directory while the directory is meant to be a python module.

I think there will be a lot of people running haikuporter directly from the git repository, but they can probably handle some reorganization.

I agree that renaming everything to pep8 is a good idea, but we can still keep the "haikuporter" name :)

jurgenwigg:
We're all on the same page :) my idea was to preserve whole functionality as it was before but in modern python package with all advantages of newer version and with possibility to release it on pypi :)

jurgenwigg:
After migration HaikuPorter seems to work as before.

Importing module in the python itself:
```shell
» python
Python 3.11.4 (main, Jun  7 2023, 12:45:49) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import HaikuPorter
>>> import haikuporter
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'haikuporter'
>>> import haiku_porter
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'haiku_porter'
>>>
```

Executing `haikuporter` script from CLI:
```shell
» ./haikuporter mesa -j4
Error: Unable to find haikuports.conf in known search paths.
Error: See haikuports-sample.conf for more information
```

Migration to **src** based structure instead of **flat** one requires installation of HaikuPorter package with `pip install .` or `pip install -e .`.

waddlesplash:
Haiku itself must be built on a case-sensitive system. You can make a case-sensitive disk in MacOS to check the repository out on and work on it that way.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
```
waddlesplash, line 1: What's all this about?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,151 @@
+============
```
waddlesplash, line 1: I think changes to README files should be done separately from such a refactor like this.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+# syntax=docker/dockerfile:1
```
waddlesplash, line 1: This file doesn't seem to have much purpose, perhaps it can be removed? BuildMaster has its own docker setup.

jurgenwigg: At this moment I agree that this file can be removed. Maybe in the future it would be helpful for developing HaikuPorter, i.e. some integration tests.

CONTRIBUTING.rst
jurgenwigg: Good point - this file should be added in separate PR. I think it would be really good to have such guide how to contribute to this project.

.pre-commit-config.yaml
jurgenwigg: It came by default with cookie cutter template which I used to create/update this python package. It depends on if we would like to have git pre-commit hooks with formatting or not. Also basic checks are included in this step. In theory it should help to keep good-looking python code in terms of style before committing it to the repo.

jurgenwigg:
> 

The question is if HaikuPorter should be created in such way that it require case sensitive system? Proposed solution should work as previous package also in terms of case sensitive/insensitive. I don't see any obstacles in having `haikuporter` script 'as is' in root directory and `HaikuPorter` directory in `src` place.

jurgenwigg:
Ok, I've decided that this PR should cover only updating existing file structure to be able to install and create proper modern python package. Everything should work like before the update. Everything related to updating documentation (use `mkdocs` or `sphinx`), CI and pre-commit hooks should be done under separate PRs.

waddlesplash:
> The question is if HaikuPorter should be created in such way that it require case sensitive system? 

Using HaikuPorter for cross-compilation requires a Haiku build tree, which needs a case-sensitive filesystem. Using HaikuPorter on Haiku will always have a case-sensitive filesystem. So there's no reason to support case-insensitive ones.

pulkomandy:
But it's just moving a few files to an src/ subdirectory, so there is not really any problem with doing it anyways. And being able to checkout the git repository on a case insensitive filesystem sounds useful for many reasons (archiving, for example).

waddlesplash:
It's not just moving "a few files" but the entire sources, and furthermore due to the changes the renames aren't even tracked so we are losing all the history here.

I would prefer the `haikuporter` command/script be moved elsewhere (1 file) than all the others.

jurgenwigg:
> It's not just moving "a few files" but the entire sources, and furthermore due to the changes the renames aren't even tracked so we are losing all the history here.
> 
> I would prefer the `haikuporter` command/script be moved elsewhere (1 file) than all the others.

The easiest way is just to add `.py` extension to existing `haikuporter` script. This is the easiest way to keep original directory for source files (`HaikuPorter`) and keep script file for it in the root directory. If this proposition would be a good deal for now? :)

waddlesplash:
Sounds fine.

jurgenwigg:
Now everything should work correctly and package itself should allow to install it through `pip install .` or `pip install -e .` command.

jurgenwigg:
@waddlesplash @scottmc please check if your comments and findings were solved and if this PR can be merged :)

.gitignore
```diff
@@ -1,3 +1,167 @@
-*.pyc
-*.pyo
-recipeCache/
+docs/source
+
+# From https://raw.githubusercontent.com/github/gitignore/main/Python.gitignore
+
+ Byte-compiled / optimized / DLL files
```
pulkomandy, line 5: I think this is missing a # 

Also probably some parts of this file are not needed.

Makefile
```diff
@@ -0,0 +1,56 @@
+.PHONY: install
+install: ## Install the poetry environment and install the pre-commit hooks
+	@echo "🚀 Creating virtual environment using pyenv and poetry"
```
pulkomandy, line 3: I think emojis are not well handled by Haiku terminal currently, maybe they should be avoided

README.md
```diff
@@ -1,23 +1,31 @@
-# Haikuporter
+# haikuporter
 
-The HaikuPorter tool is provided to ease the fetching, patching and building of source code. It can be compared to a slim version of [Gentoo Portage](https://www.gentoo.org/main/en/about.xml). Each port contains the [Haiku](http://haiku-os.org) specific patches to the original source code. It fetches the original source code, applies the Haiku-specific patches, builds the software, and packages it.
```
pulkomandy, line 3: this part of the README seems quite important, explaining what the tool does.

On the other hand I don't think the collection of "shields" is really that useful.

README.md
```diff
@@ -70,3 +78,7 @@ erase the container without losing your work.
  - useradd sshd ; ssh-keygen -A
  - Enable PermitRootLogin in /system/settings/ssh/sshd\_config and make sure the path to the sftp server is correct
  - install xz\_utils\_x86, lzip\_x86 (required for extracting packages), patch, dos2unix (required for PATCH() function in some packages)
+
+---
+
+Repository initiated with [fpgmaas/cookiecutter-poetry](https://github.com/fpgmaas/cookiecutter-poetry).
```
pulkomandy, line 84: Can we do without that advertising clause?

poetry.lock
```diff
@@ -0,0 +1,1340 @@
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
pulkomandy, line 1: do we really need this 1000-line files to declare dependencies? I may not be up to date with the current Python approach, but wouldn't a simple requirements.txt for use with pip be enough?

jurgenwigg: I think we don't need this file, it had been uploaded by accident.

README.md
jurgenwigg: Since we heavily modded original template we can remove this clause.

README.md
jurgenwigg: I agree - this is definitely more useful than shields. I'm changing this.

jurgenwigg: Done

poetry.lock
jurgenwigg: Deleted.

README.md
jurgenwigg: Done

Makefile
waddlesplash: Seems like not, it can be removed?

.gitignore
```diff
@@ -1,3 +1,146 @@
...
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+#.idea/.history
+
+.history*
+.vscode*
+poetry.lock
 *.pyc
 *.pyo
 recipeCache/
```
waddlesplash, line 146: probably would be good to keep HaikuPorter-specific items at the top

setup.py
```diff
@@ -1,29 +0,0 @@
-"""Setup script for HaikuPorter
```
waddlesplash, line 1: I think we actually use this to package HaikuPorter? Not sure...

jurgenwigg: setup.py is not needed and can be safely replaced by newer solution which is using pyproject.toml to install the package. Installing package with pip works flawlessly. I'll link article/document describing why setup.py is deprecated and how should it be replaced.

pyproject.toml
```diff
@@ -0,0 +1,102 @@
+[tool.poetry]
```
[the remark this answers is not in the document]
jurgenwigg: Poetry is like cargo in rust. You don't have to use it, but I treat it like a "touch of the future" where python has better package manager than pip ;)

.gitignore
jurgenwigg: Done

Makefile
jurgenwigg: Done

setup.py
jurgenwigg: And the links: 
1. https://peps.python.org/pep-0517/
2. https://peps.python.org/pep-0518/
3. https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary

Begasus:
Thanks for working on this, first problem arizes when running the checks on haikuports, it can't find `haikuporter` (as it's renamed it's  no wonder but that should be addressed). :)
My guess is that if haikuporter gets a new release buildmaster(s) will be affected also by this? (eg haikuporter vs haikuporter.py in the conf file)

waddlesplash:
It may make sense to restore the old filename, then...

jurgenwigg:
I'm back from the weekend - I'm checking what went wrong with changing the filename.

jurgenwigg:
#262 I've created PR for fixing this issue.

PR #262, opened 2023-07-23
[FIX] Fix problems caused by changing script file name

jurgenwigg:
While making the package modern I've changed filename of `haikuporter` to `haikuporter.py`. I didn't notice that this script is included in other places and is actively used by the scripts. I've changed all occurrences of script file name that is appearing in other files.

### DIFF
--- a/buildmaster/backend/assets/bin/missing_ports.sh
+++ b/buildmaster/backend/assets/bin/missing_ports.sh
@@ -10,7 +10,7 @@ fi
 
 set -e
 
-HAIKUPORTER=${HAIKUPORTER:-haikuporter}
+HAIKUPORTER=${HAIKUPORTER:-haikuporter.py}
 VALID_PACKAGE_LIST=$(mktemp)
 PRESENT_PACKAGE_LIST=$(mktemp)
 
--- a/doc/haikuporter_bash_completion
+++ b/doc/haikuporter_bash_completion
@@ -41,4 +41,4 @@ function _haikuporter()
 
 	return 0
 }
-complete -F _haikuporter -o filenames haikuporter
+complete -F _haikuporter.py -o filenames haikuporter
--- a/doc/haikuporter_bash_completion
+++ b/doc/haikuporter_bash_completion
@@ -5,7 +5,7 @@
 
 function _hp_options
 {
-local options="$(haikuporter -h | grep -- -- | cut -f3,4 -d' ' | sed -e 's/, / /')"
+local options="$(haikuporter.py -h | grep -- -- | cut -f3,4 -d' ' | sed -e 's/, / /')"
 COMPREPLY=(${COMPREPLY[@]:-} $(compgen -W '$options' -- "$cur"))
 }
 
@@ -13,7 +13,7 @@ function _hp_beps
 {
 	local packages
 
-	for i in "$(haikuporter -t)"/*/*/*.bep; do
+	for i in "$(haikuporter.py -t)"/*/*/*.bep; do
 		i="${i##*/}"
 		packages="${i%.bep} ${packages}"
 	done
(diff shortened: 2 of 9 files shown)


jurgenwigg:
To make this PR better and to introduce first tests - could you provide some potential use cases of `haikuporter.py` script for me to create and implement integration tests for this part of the package?

buildmaster/backend/assets/bin/missing_ports.sh
```diff
@@ -10,7 +10,7 @@ fi
 
 set -e
 
-HAIKUPORTER=${HAIKUPORTER:-haikuporter}
```
jurgenwigg, line 13: I'm not sure how this variable in `sh` script works. I need to check if it is ok now.

doc/haikuporter_bash_completion
```diff
@@ -41,4 +41,4 @@ function _haikuporter()
 
 	return 0
 }
-complete -F _haikuporter -o filenames haikuporter
```
jurgenwigg, line 44: Similar thing as in the previous comment - need to figure out if this change is ok or not.

jurgenwigg: Nów this should be fixed, but please check if I'm fixed it correctly.

buildmaster/backend/assets/bin/missing_ports.sh
jurgenwigg: According to https://unix.stackexchange.com/questions/122845/using-a-b-for-variable-assignment-in-scripts it looks like this change is ok, but please check it.

waddlesplash:
I think it would probably be better to just add the symlink back, even if this means the repo can't be checked out on case-insensitive filesystems.

jurgenwigg:
I think we should try to fix/address the issue before we will restore previous file name. It looks that all places where original script was called I've changed.

waddlesplash:
But is there really any reason to change the filename, besides case-insensitive filesystem compatibility? As previously mentioned, Haiku requires case-sensitive filesystems even for cross-compilation.

There are places outside this repository where the name sans `.py` are referenced, too.

pulkomandy:
We already had this discussion in the previous merge request. Being able to checkout the repository on any OS is a good enough reason.

jurgenwigg:
@OscarL Totally agree - I didn't tick the checkbox for the option for squashing the commits... Maybe setting this option as default (encourage to squash commits while merging the PR with ticked checkbox) would be a good idea?

jurgenwigg:
@OscarL GitHub, like GitLab has the option to tick the checkbox for squashing commits while merging PR by default can be set only by maintainers (in GitLab nomenclature) so we can only to check that box manually BEFORE doing the merge action. Pushing and using git itself is not related to squashing commits while merging and handling PR itself by the GitHub.

After merging the PR it's a lot harder to make a squash. It needs to make the rebase to the commit before the merge and 'recreate' the history with squashing the commits.

I was looking for a good project to contribute to for some time. And then Haiku appeared (again) and it turns out that this project is alive and it keeps going so I've decided to start contributing :)

jurgenwigg:
Let me explain what happened here:
1. I was waiting for the green light to merge this PR. 
2. PR was merged and I noticed that after it was made.
3. Yes, you are right - my commit messages are not as good as they should, but I was prepared for squashing them under one commit message.
4. I think I should prepare commit message while creating the PR itself.

For me this piece of documentation is missing - what is the workflow for contributing here, who is merging the PR, etc. 

I agree with your arguments @OscarL. The question is who should rewrite/squash this commit history. I can do that but I think I don't have such permissions.

pulkomandy:
I did the review on my phone and forgot to switch to "squash and merge", which happens everytime I use the github user interface because it does everything it can to hide the commit messages when reviewing.

No need for 10 messages debating about it.

PR #263, opened 2023-07-24
Formatted with black and isort

jurgenwigg:
Formatted code with black and isort. Fixed found small issue in `MockBuilder.py`.

### DIFF
--- a/HaikuPorter/BuildMaster.py
+++ b/HaikuPorter/BuildMaster.py
@@ -7,18 +7,6 @@
 
 # -- Modules ------------------------------------------------------------------
 
-from .ConfigParser import ConfigParser
-from .Configuration import Configuration
-from .Options import getOption
-from .Port import Port
-from .ReporterMongo import ReporterMongo
-from .ReporterJson import ReporterJson
-from .Utils import ensureCommandIsAvailable, sysExit, warn, info
-
-from .Builders.Builder import _BuilderState
-from .Builders.LocalBuilder import LocalBuilder
-from .Builders.RemoteBuilderSSH import RemoteBuilderSSH
-
 import errno
 import json
 import logging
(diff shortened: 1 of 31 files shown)


waddlesplash (changes requested):
I definitely do not agree with switching to spaces for indent.

I know that's general Python convention, but Haiku's has always been to use tabs. Python itself tolerates this just fine; there's no real reason to switch now. It will just make blames more difficult, and introduces churn for no good reason, IMO.

jurgenwigg:
From my perspective and experience 4 spaces are a lot better. Space is very good specified - white character that has one letter long. Tab has various length depending on the system where is present. 

Since there is no Coding Guideline that is applicable to Python I suggest to use PEP8 which is widely used in Python community. 

`black` is using spaces for indents without setting anything. Changing to tabs will require passing parameter for that.

I'll promote spaces for better interoperability and better consistency instead of using tabs. But in the end the decision of using tabs instead of spaces should be a decision of community not a single person. If tabs will be chosen as a preferred indents then some kind of Coding Guidelines for Python should be created.

Begasus:
If this would take it's turn on the recipes I'm for keeping the tabs (call me old-fasion) :)
For pure python I don't' know as I'm not familiar there.

jurgenwigg:
Discussion about using tabs/spaces for indents here affects python files/code only :)

pulkomandy:
Everything in Haiku is indented with tabs, it will be hard to change our minds on this even if it is nonstandard in the Python world.

jurgenwigg:
So I have more general question - do we want to apply C-related Coding Guideline to the Python code? PEP8 also suggests different naming style for methods or variables defined. 

I'm aware that PEP8 in terms of spaces vs tabs allows using tabs in case when existing repository is using tabs, but personally I think that in such case there should be proper Coding Guideline that defines it.

Many tools is able to make formatting and linting the code according to PEP8 out-of-the-box and applying some custom rules means also some additional configuration for them.

waddlesplash:
The Coding Guidelines for C++ apply to all programming languages insofar as they can. Yes, there are some guidelines that don't make sense or don't apply to Python code (or Bash or Perl or JavaScript or ...) but all the ones that can apply should be followed.

PR #265, opened 2023-07-25
[DRAFT] Formatted files with yapf

jurgenwigg:
Formatted python files with yapf.
Commands used:
```shell
$ yapf -r --style='{based_on_style:pep8,use_tabs:True,column_limit=88}' --in-place haikuporter.py
$ yapf -r --style='{based_on_style:pep8,use_tabs:True,column_limit=88}' --in-place HaikuPorter/
```

Base on **PEP8**. Key differences from **PEP8**:
- Line length set to 88 characters instead of 79
- Used tabs instead of 4 spaces

### DIFF
--- a/HaikuPorter/BuildMaster.py
+++ b/HaikuPorter/BuildMaster.py
@@ -29,13 +29,16 @@ import subprocess
 import threading
 import time
 
+
 class ThreadFilter(object):
+
 	def __init__(self):
 		self.ident = threading.current_thread().ident
 		self.build = None
 
 	def reset(self):
 		self.ident = threading.current_thread().ident
+
 	def setBuild(self, build):
 		self.build = build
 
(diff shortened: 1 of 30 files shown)


waddlesplash (changes requested):
I still see various problems here:

 * Continuation lines are indented with spaces, not tabs.
 * Comments on the next line should be indented, this was lost.
 * Operators were moved to the previous line instead of the next for continuations.

So, I don't see how this adds any value.

jurgenwigg:
I've noticed the same issues - changing title to draft as this is under development to address and fix all issues.

waddlesplash:
But what changes does this actually produce, then? Most of the differences are due to those 3 problems I noted. Once they are corrected, what "differences" will remain?

I think the code is formatted mostly fine as-is, and we don't need to reformat it "just because".

jurgenwigg:
Main goal was to prepare codebase for linters and implementing CI for checking imports order, basic styling/formatting of the code for each new PR. I'm also trying to figure out which tool and how can be used by other users to quickly format the code before committing any changes.

waddlesplash:
I don't know if we can use machine-formatted code, due to the second item listed above at least (comments on next lines should be indented.) There's probably others (e.g. I see various multiline object initializers were condensed to one line.)

jurgenwigg:
I'll try to create a draft document with coding guidelines and tools possible to use for formatting :) I'll take Haiku Coding Guidelines as a base.

waddlesplash:
Do we really need a separate document, though?

An "addendum" or "supplement" document describing only the differences would likely make much more sense.

jurgenwigg:
@waddlesplash I had exactly that on my mind - some kind of appendix to existing documentation. "Documentation" sounds pretty serious but all in all every comment is in some way a kind of documentation ;)

jurgenwigg:
This PR is almost year old. I'm closing it since another approach was proposed in the comments.

Issue #268, opened 2023-07-27
Cannot clone repository on MacOS


jurgenwigg:
Hi, after last commits I cannot clone repository on MacOS:
```shell
$ git clone git@github.com:jurgenwigg/haikuporter.git
Cloning into 'haikuporter'...
remote: Enumerating objects: 6334, done.
remote: Counting objects: 100% (700/700), done.
remote: Compressing objects: 100% (333/333), done.
remote: Total 6334 (delta 419), reused 586 (delta 354), pack-reused 5634
Receiving objects: 100% (6334/6334), 1.86 MiB | 504.00 KiB/s, done.
Resolving deltas: 100% (4523/4523), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'haikuporter'
```

waddlesplash:
You can clone the repository by creating a case-sensitive filesystem image, the same as for building Haiku. Here's an old guide that details how to do it: https://www.haiku-os.org/articles/2015-02-05_building_haiku_mac_os_x_1010_yosemite/#disk_image

pulkomandy:
@jurgenwigg made some simple renaming changes to fix this and they were cancelled by a later change.

I think we should really fix this repository to be cloneable without such complicated setup. Why is that so hard to do? It's just about renaming a single file :(

jurgenwigg:
We can also change path to the source code by just moving files from `HaikuPorter` directory to the `src/HaikuPorter`. Some minor fixes in the imports inside, but external API (script file name) should be the same. The best way would be to suggest and promote approach to just install the `HaikuPorter` package in the system with `pip` command and use it as a normal OS command to build the packages.

waddlesplash:
The reversion was done by @korli.

> to just install the HaikuPorter package in the system with pip command 

But HaikuPorter does have real usages outside of Haiku, namely bootstrap cross-compile builds, which are not done with installed toolchains. That would be the primary usecase for HaikuPorter on macOS. As the Haiku repository itself cannot be checked out at all, much less built, on a case-insensitive filesystem, why does it matter that HaikuPorter can't be, either?

jurgenwigg:
There are many use cases and reasons to make possible for the people to normally (without any issues) clone repository. Before doing any further refactoring or cleanup we should write unit and integration test for the existing code. This should be possible to be done on any OS. 

[...]

PR #269, opened 2023-07-31
[DRAFT] Move source code to src directory

jurgenwigg:
Proposition of moving source code of `HaikuPorter` to the `src` directory. This PR also includes some small changes that creates OS command `haikuporter` after installation with `pip`. 

This is a draft - there is need to make changes in various places and update `README` according to introducing the changes. 

#268 This a proposition to solve this problem.


jurgenwigg:
This PR is already almost year old. I'm closing it due to various discussions.

PR #275, opened 2023-10-10
If `pkgman install` fails, let the user see what the problems were.

### LINKED ISSUE #256: Explicitly state which dependency could not be satisfied (or why)
While trying to build "opencolorio-2.0.1" on beta4, 32 bits (x86), I get:

```
Fetching package for devel:libboost_system_x86 >= 1.70.0 ...
Fetching package for devel:libglew_x86 >= 2.2 ...
Error: unable to resolve required packages for build for opencolorio_x86-2.0.1
Error:  dependency-infos:
Error:          /boot/home/SourceCode/haikuports/haikuports/repository/opencolorio_x86-2.0.1.DependencyInfo
Error:          /boot/home/SourceCode/haikuports/haikuports/repository/opencolorio_x86_devel-2.0.1.DependencyInfo
Error:  repositories:
Error:          ['/boot/home/SourceCode/haikuports/haikuports/packages', '/boot/system/packages']
Error: 
```

It would be nice if the package that failed to be found/install was more clearly stated, instead of having to do a manual pkgman search for each of the dependencies.

OscarL:
Fix #256.

### DIFF
--- a/HaikuPorter/DependencyResolver.py
+++ b/HaikuPorter/DependencyResolver.py
@@ -7,7 +7,7 @@
 
 import os
 import re
-from subprocess import CalledProcessError, check_output
+import subprocess
 
 from .Options import getOption
 from .PackageInfo import PackageInfo, ResolvableExpression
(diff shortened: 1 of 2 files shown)


HaikuPorter/DependencyResolver.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 import re
-from subprocess import CalledProcessError, check_output
+from subprocess import CalledProcessError, check_output, PIPE
```
waddlesplash, line 10: Probably we shouldn't pollute global namespace, we should just import subprocess and then use it as `subprocess.PIPE`.

HaikuPorter/DependencyResolver.py
```diff
@@ -238,15 +238,25 @@ def _addImmediate(self, parent, requires, typeString, forBuildhost):
 			if getOption('getDependencies'):
 				try:
 					print('Fetching package for ' + str(requires) + ' ...')
-					output = check_output(['pkgman', 'install', '-y', str(requires).replace(' ', '')]).decode('utf-8')
+					output = check_output(['pkgman', 'install', '-y', str(requires).replace(' ', '')], stderr=PIPE).decode('utf-8')
```
waddlesplash, line 241: line is too long.

PR #276, opened 2023-10-20
ShellScriplets.py: Create addGamesDeskbarSymlink

erysdren:
I thought it would be interesting to create a scriplet for adding an application to a "Games" category on the Deskbar. Seems odd to me that such a thing isn't already in.

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -695,6 +695,32 @@ addAppletDeskbarSymlink()
 	symlinkRelative -s "$appPath" "$targetDir/$entryName"
 }
 
+addGamesDeskbarSymlink()
+{
+	# Usage: addGamesDeskbarSymlink <appPath> [ <entryName> ]
+	# Creates a Deskbar symlink for a game application.
+	# <appPath> is the absolute path to the application executable.
+	# <entryName> is the name of the application as it shall appear in the
+	# Deskbar. Can be omitted, in which case the name of the application
+	# executable is used.
+	if [ $# -lt 1 ]; then
+		echo >&2 "Usage: addGamesDeskbarSymlink <appPath> [ <entryName> ]"
+		exit 1
+	fi
+
+	appPath="$1"
+	shift 1
+	if [ $# -lt 1 ]; then
+		entryName="$(basename "$appPath")"
+	else
+		entryName=$1
+	fi
+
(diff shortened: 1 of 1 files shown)


coolcoder613eb:
Next up: Development?

erysdren:
i agree with that. would that be a better change?

humdingerb:
This boils down to the old discussion on how to auto-categorize the entries in Deskbar, check the forums, it comes up from time to time.

As is, I don't think this solution is that nice.
* It forces a "Games" sub-folder on everyone. Maybe I like opening the "Applications" folder and use type-ahead-filtering to find my apps. Putting stuff in subfolders sabotages that.
* The "Games" folder isn't translatable.
* If it doesn't stay with the one "Games" folder but becomes a general categorization, the above issues multiply. And what to do if things fit several categories or I look in "Productivity" (whatever that is) and it's in "Office" or whatever.

One possible solution that has been discussed (and at least partly implemented as an experiment by mmu_man IIRC) is only add a category attribute and leave it to the user to use a query-system to do the sorting. Has to be wired into Deskbar + settings yet.

Anyway, I don't think forcing this subfolder is a good idea.

humdingerb:
I point once more to the Deskbar / categories threads in the forum to avoid rehashing it here all again.

The gist is, categorization should be optional and the categories localized and user-editable. To create precedents by just going ahead and force the categories as subfolder on everyone isn't OK and only revertable by re-building a ton of packages. A means to deal with categories needs to be implemented in Deskbar first.

pulkomandy:
There are no short term or even medium term plans to actually implement this. So, uuntil then, I think a Games category is nt a controversial one (there is a pretty clear distinction of what is a game and what isn't).

So I think we should goahead with this specific one for the time being.

When we implement some other way to populate DeskBar, it will require rebuilding all packages anyways, to add needed metadata in the new format. So I don't see what we win by not moving games to a separate directory?

humdingerb:
Alright then. Now that people see there's a Game category, we'll see if there's demand for more...

mmuman:
How about a more generic AddCategoryDeskbarSymlink that gets the category passed, which for now just handles Games separately?
This way we wouldn't have to patch recipes if we change our minds on how it's laid out, either directories or attributes…

At one point I thought of using the recipe location as a hint (app-*, games-*…) to categorize things automatically, or at least by default.

PR #281, opened 2024-02-23
Add `cmd:find` to scriptletPrerequirements (needed by fixCMake).

OscarL:


HaikuPorter/ShellScriptlets.py
```diff
@@ -15,6 +15,7 @@
 	coreutils
 	cmd:bash
 	cmd:grep
+	cmd:find
```
jmairboeck, line 18: This should come before `cmd:grep`, I suppose.

PR #285, opened 2024-04-10
don't add the secondaryArchSuffix on patchset export

TheZeldakatze:
haikuporter always generated package_x86.patchset files.
Now it should always use the basename so you don't have to manually rename the patchset files

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -711,14 +711,14 @@ class Port(object):
 		## use enumerate to avoid your own couters
 		for s, source in enumerate(self.sources, 1):
 			if s == 1:
-				patchSetFileName = self.name + '-' + self.version + '.patchset'
-				archPatchSetFileName = (self.name + '-' + self.version + '-'
+				patchSetFileName = self.baseName + '-' + self.version + '.patchset'
+				archPatchSetFileName = (self.baseName + '-' + self.version + '-'
 										+ self.targetArchitecture
 										+ '.patchset')
 			else:
-				patchSetFileName = (self.name + '-' + self.version + '-source'
+				patchSetFileName = (self.baseName + '-' + self.version + '-source'
 									+ str(s) + '.patchset')
-				archPatchSetFileName = (self.name + '-' + self.version + '-'
+				archPatchSetFileName = (self.baseName + '-' + self.version + '-'
 										+ self.targetArchitecture + '-source'
 										+ str(s) + '.patchset')
 			patchSetFilePath = self.patchesDir + '/' + patchSetFileName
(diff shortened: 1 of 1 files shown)


Begasus:
Could be handy, manually changing it in 32bit secondary architecture has always been a pain, needs testing.

HaikuPorter/Port.py
```diff
@@ -711,14 +711,14 @@ def extractPatchset(self):
 		## use enumerate to avoid your own couters
 		for s, source in enumerate(self.sources, 1):
 			if s == 1:
-				patchSetFileName = self.name + '-' + self.version + '.patchset'
-				archPatchSetFileName = (self.name + '-' + self.version + '-'
+				patchSetFileName = self.baseName + '-' + self.version + '.patchset'
+				archPatchSetFileName = (self.baseName + '-' + self.version + '-'
```
jmairboeck, line 715: Is `archPatchSetFileName` supposed to be for architecture specific patches? I haven't actually looked at what this does exactly, but if that is the case, maybe the original name should stay for these?

In a recipe, `$secondaryArchSuffix` could be used to refer to them, if needed.

Begasus: A foo$secondaryArchSuffix.patchset isn't the way I think we should go, the intention should be to use the patchset over the architectures.
I'm not familiar with the `archPatchSetFileName` function, but did a check and it's working as expected.

jmairboeck: I meant to only refer to the second assignment here, the first line is fine, I think. I was just speculating what the second one could be for.

Begasus: Ah! my bad :) Maybe someone with some more insight can shed a light here?

Begasus: Checking `patchSetFileName` only results in the expected outcome, checking `archPatchSetFileName` only results in the _x86 suffix still for the patchset, so that shouldn't bee used.

HaikuPorter/Port.py
```diff
@@ -711,14 +711,14 @@ def extractPatchset(self):
...
 										+ self.targetArchitecture
 										+ '.patchset')
 			else:
-				patchSetFileName = (self.name + '-' + self.version + '-source'
+				patchSetFileName = (self.baseName + '-' + self.version + '-source'
 									+ str(s) + '.patchset')
-				archPatchSetFileName = (self.name + '-' + self.version + '-'
+				archPatchSetFileName = (self.baseName + '-' + self.version + '-'
```
pulkomandy, line 721:
```suggestion
				archPatchSetFileName = (self.name + '-' + self.version + '-'
```

PR #288, opened 2024-07-01
Unit tests for RecipeTypes.py

jurgenwigg:
I've created unit tests for the `RecipeTypes` module. This is a first step before making any refactors.

### DIFF
--- a/tests/test_recipetypes.py
+++ b/tests/test_recipetypes.py
@@ -1,0 +1,106 @@
+"""Unit tests for RecipeTypes.py"""
+from HaikuPorter.RecipeTypes import (Architectures, MachineArchitecture, Phase,
+                                     YesNo)
+from pytest import mark
+
+_ARCHITECTURE_LIST = [
+    "arm",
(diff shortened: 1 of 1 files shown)


waddlesplash:
These tests don't seem very interesting, they just duplicate a bunch of constants and assert that they are the same.

jurgenwigg:
Agree, but I prefer to write even that simple tests before making refactor in the file.

waddlesplash:
But if all the test file does is copy data from the file, then it's not really worth anything, but just makes one more thing you have to update whenever modifying the file.

jurgenwigg:
In case when there are no unit tests at all and some code depends on the data and API of `RecipeTypes` I prefer to have UT for testing the API. Generally you are right, but I would like to refactor `RecipeTypes` as it is stated in the comments in the code and have some form of source of truth.

waddlesplash:
RecipeTypes is already the "source of truth" for these values, though. The only other possible source would be the Haiku source code.

PR #291, opened 2024-08-28
Implement S3 Object Storage for Package Repositories

### LINKED ISSUE #258: buildmaster needs to leverage object storage
Constraints:
  * 6 volume mounts per node (OS + 6)
    * https://docs.digitalocean.com/products/volumes/details/limits/
  * Pods that share volumes must schedule on the same node 
    * https://docs.digitalocean.com/products/volumes/details/limits/

Considerations:
  * Persistent volumes are expensive.  Object storage is cheap

[Volumes today:](https://github.com/haiku/infrastructure/blob/master/deployments/buildmaster.yml)
  * buildmaster-sources (shared by every buildmaster)
  * buildmaster-packages (shared by every buildmaster and frontend)
  * buildmaster-x86-gcc2
  * buildmaster-x86-64

We might be able to combine the buildmaster work volumes into one big shared pool of disk... but it really starts to shift this stuff back to a big pile of spaghetti.   (however, this doesn't solve the high cost of direct attach storage either)

mmlr:
* Adds object storage based backing for the package repository (S3 API only for now)
* Reduces the number of volumes that are used to ease deployment
* Fixes constant reuploads of system-packages to builders
* Other minor fixes and cleanups

After deployment this should fix #258.

### DIFF
--- a/HaikuPorter/BuildMaster.py
+++ b/HaikuPorter/BuildMaster.py
@@ -43,7 +43,7 @@ class ThreadFilter(object):
 
 
 class ScheduledBuild(object):
-	def __init__(self, port, portsTreePath, requiredPackageIDs,
+	def __init__(self, port, portsTreePath, missingPackageIDs,
 		packageRepository, presentDependencyPackages):
 		self.port = port
 		self.recipeFilePath \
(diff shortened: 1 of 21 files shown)


kallisti5 (approved):
Nice idea!  The licenses were always a pain.

PR #294, opened 2024-09-02
HaikuPorter/ShellScriptlet.py: Fixed infinite recursion

### LINKED ISSUE #293: Bootstrap builds broken since commit 35a32ad20a9f396625c7e176f7bac155f3e71378
It seems this commit broke bootstrap builds. Didn't investigate that much further but when trying to build a ppc or arm64 bootsteap image it gets to building zlib_bootstrap package and when building it it seems that the wrapper script gets into an infinite recursion when trying to find cmake until it is killed / segfaults. After reverting this commit it works again.

I think the problem is that the function is called cmake and command -v will return that function name instead of searching the cmake binary. Below is a small test case sctipt, try it with bash -x and you'll see that it recurses infinitely, but if you comment out the command -v line and uncomment the which line it finds the cmake and excutes it.

```
#!/bin/bash

cmake()
{
	echo "start"
	CMAKE="/temp/cmake"
	if [ ! -f $CMAKE ]; then
		CMAKE=$(command -v cmake)
		#CMAKE=$(which cmake)
		echo "command -v returns \"$CMAKE\""
	fi
	echo "CMAKE = $CMAKE"
	$CMAKE 
	echo "end"
}

cmake

```

ilzu:
…on bootstrap builds.

Fixes #293

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -332,7 +332,6 @@ cmake()
 	CMAKE=$portPackageLinksDir/cmd~cmake/bin/cmake
 	if [ ! -f $CMAKE ]; then
 		CMAKE=$(type -Pp cmake)
-        echo "CMAKE = $CMAKE"
 	fi
 
 	$CMAKE "$@"
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -331,7 +331,8 @@ cmake()
 
 	CMAKE=$portPackageLinksDir/cmd~cmake/bin/cmake
 	if [ ! -f $CMAKE ]; then
-		CMAKE=$(command -v cmake)
+		CMAKE=$(type -Pp cmake)
+        echo "CMAKE = $CMAKE"
 	fi
 
 	$CMAKE "$@"
@@ -348,7 +349,7 @@ meson()
 
 	MESON=$portPackageLinksDir/cmd~meson/bin/meson
 	if [ ! -f $MESON ]; then
-		MESON=$(command -v meson)
+		MESON=$(type -Pp meson)
 	fi
 
 	$MESON --wrap-mode=nodownload "$@"


HaikuPorter/ShellScriptlets.py
```diff
@@ -331,7 +331,8 @@ def getShellVariableSetters(shellVariables):
 
 	CMAKE=$portPackageLinksDir/cmd~cmake/bin/cmake
 	if [ ! -f $CMAKE ]; then
-		CMAKE=$(command -v cmake)
+		CMAKE=$(type -Pp cmake)
```
jmairboeck, line 334:
```suggestion
		CMAKE=$(type -P cmake)
```

HaikuPorter/ShellScriptlets.py
```diff
@@ -348,7 +349,7 @@ def getShellVariableSetters(shellVariables):
 
 	MESON=$portPackageLinksDir/cmd~meson/bin/meson
 	if [ ! -f $MESON ]; then
-		MESON=$(command -v meson)
+		MESON=$(type -Pp meson)
```
jmairboeck, line 352:
```suggestion
		MESON=$(type -P meson)
```

HaikuPorter/ShellScriptlets.py
```diff
@@ -331,7 +331,8 @@ def getShellVariableSetters(shellVariables):
 
 	CMAKE=$portPackageLinksDir/cmd~cmake/bin/cmake
 	if [ ! -f $CMAKE ]; then
-		CMAKE=$(command -v cmake)
+		CMAKE=$(type -Pp cmake)
+        echo "CMAKE = $CMAKE"
```
jmairboeck, line 335: Maybe remove this debug output?

jmairboeck (approved):

HaikuPorter/ShellScriptlets.py
ilzu: Damn, I forgot that there.

HaikuPorter/ShellScriptlets.py
ilzu: I think it needs both -P (search only in path, not in functions and aliases) and -p (just return the file name, or nothing)

ilzu: perhaps even -fPp (f = suppress shell function lookup)

jmairboeck: It seems to work fine `-Pp`. The extra `p` is no big deal. Thanks for fix!

PR #296, opened 2024-09-05
Source/SourceFetcher: let the fetcher compute the checksum

korli:
* checksums for git commits or git tags can then be enabled, example: git+https://github.com/HaikuArchives/StreamRadio.git#tag=1.0.0 or #commit=$srcGitRev
* this works by hashing the checkout content. inspired by the same feature in pacman from Arch.

### DIFF
--- a/HaikuPorter/SourceFetcher.py
+++ b/HaikuPorter/SourceFetcher.py
@@ -12,6 +12,7 @@
 
 # -- Modules ------------------------------------------------------------------
 
+import hashlib
 import os
 import re
 import shutil
@@ -104,6 +105,19 @@ def foldSubdirIntoSourceDir(subdir, sourceDir):
 	for fileName in os.listdir(fullSubdirPath):
 		os.rename(fullSubdirPath + '/' + fileName, sourceDir + '/' + fileName)
 	os.removedirs(fullSubdirPath)
+
+# -----------------------------------------------------------------------------
+
+def calcChecksumFile(file):
+	sha256 = hashlib.sha256()
+
+	with open(file, 'rb') as f:
+		while True:
+			data = f.read(16384)
+			if not data:
+				break
+			sha256.update(data)
+	return sha256.hexdigest()
 
 # -- Fetches sources via bzr --------------------------------------------------
 
@@ -227,6 +241,9 @@ class SourceFetcherForDownload(object):
 		unpackFile(self.uri, self.fetchTarget, sourceBaseDir, sourceSubDir,
 			foldSubDir)
 
+	def calcChecksum(self):
+		return calcChecksumFile(self.fetchTarget)
+
 # -- Fetches sources via fossil -----------------------------------------------
 
 class SourceFetcherForFossil(object):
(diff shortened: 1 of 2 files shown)


HaikuPorter/SourceFetcher.py
```diff
@@ -273,19 +290,27 @@ def __init__(self, uri, fetchTarget):
...
-		warn("UNSAFE SOURCES ARE BAD AND SHOULD NOT BE USED IN PRODUCTION")
-		warn("PLEASE MOVE TO A STATIC ARCHIVE DOWNLOAD WITH CHECKSUM ASAP!")
+			warn("UNSAFE SOURCES ARE BAD AND SHOULD NOT BE USED IN PRODUCTION")
+			warn("PLEASE MOVE TO A STATIC ARCHIVE DOWNLOAD WITH CHECKSUM ASAP!")
 
 		ensureCommandIsAvailable('git')
 		command = 'git clone --bare %s %s' % (self.uri, self.fetchTarget)
 		output = check_output(command, shell=True, stderr=STDOUT).decode('utf-8')
 		info(output)
+		# sanitize the cloned repository, attributes could affect the export
```
waddlesplash, line 310: add blank line above this?

HaikuPorter/SourceFetcher.py
```diff
@@ -273,19 +290,27 @@ def __init__(self, uri, fetchTarget):
...
-					'haikuports.conf!')
+		if not self.sourceShouldBeValidated:
+			if not Configuration.shallAllowUnsafeSources():
+				sysExit('Downloading from unsafe sources is disabled in ' +
+						'haikuports.conf!')
 
-		warn("UNSAFE SOURCES ARE BAD AND SHOULD NOT BE USED IN PRODUCTION")
-		warn("PLEASE MOVE TO A STATIC ARCHIVE DOWNLOAD WITH CHECKSUM ASAP!")
+			warn("UNSAFE SOURCES ARE BAD AND SHOULD NOT BE USED IN PRODUCTION")
+			warn("PLEASE MOVE TO A STATIC ARCHIVE DOWNLOAD WITH CHECKSUM ASAP!")
```
waddlesplash, line 304: This isn't the case anymore I suppose, so we can remove all the "unsafe sources" logic? Or at least, we can remove it from this branch?

korli: Actually it would be more "PLEASE MOVE TO A TAG OR COMMIT WITH CHECKSUM!"

HaikuPorter/SourceFetcher.py
korli: Done

HaikuPorter/SourceFetcher.py
korli: Done

waddlesplash (approved):

HaikuPorter/SourceFetcher.py
```diff
@@ -273,20 +290,29 @@ def __init__(self, uri, fetchTarget):
...
 			self.rev = 'HEAD'
+		if self.rev.startswith('tag=') or self.rev.startswith('commit='):
+			self.rev=self.rev[self.rev.find('=') + 1:]
+			self.sourceShouldBeValidated = True
 
 	def fetch(self):
-		if not Configuration.shallAllowUnsafeSources():
-			sysExit('Downloading from unsafe sources is disabled in ' +
-					'haikuports.conf!')
+		if not self.sourceShouldBeValidated:
```
waddlesplash, line 298: Hmm, what happens if we have this set, but then there's no checksum? Won't it then allow the source through without validating the checksum...?

korli: When there is no checksum, this is checked at https://github.com/haikuports/haikuporter/blob/5e5a2b4b49382c84b5154e508b911fff6bda74d1/HaikuPorter/Source.py#L274

waddlesplash: OK, makes sense then.

waddlesplash (approved):

PR #302, opened 2024-10-26
add generic perl recipes

jmairboeck:
I also added some explaining comments to the main recipe.

### DIFF
--- a/generic/generic_perl-1.23.recipe
+++ b/generic/generic_perl-1.23.recipe
@@ -1,0 +1,73 @@
+SUMMARY="Short description of ProjectX"
+DESCRIPTION="Long ProjectX description.
+Spanning several paragraphs, maybe
+ - containing
+ - bullet points etc.
+You can also have really long lines that do not fit in 100 characters and use a backslash to \
+wrap the line."
+HOMEPAGE="https://metacpan.org/pod/Project::X"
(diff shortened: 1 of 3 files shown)


generic/generic_perl-1.23.recipe
```diff
@@ -0,0 +1,73 @@
...
+LICENSE="MIT
+	GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://cpan.metacpan.org/authors/id/A/AU/AUTHOR/Project-X-$portVersion.tar.gz"
+CHECKSUM_SHA256="000000000000000000000000000000000"
+SOURCE_DIR="Project-X-$portVersion"
+PATCHES="project_x-$portVersion.patchset"
+ADDITIONAL_FILES="projectx.rdef"
+
+ARCHITECTURES="any"
```
jmairboeck, line 20: Should we mark these with `?`? Most other generic recipes use it.

Begasus: I didn't really change that on the other ones, so I guess for these it would be good to also use "?"

Begasus:
Thanks, should check some perl packages with these. 👍

jmairboeck:
Now the REQUIRES are sorted correctly, sorry for that.

PR #309, opened 2024-12-29
Add support for pre-uninstall scripts

### LINKED ISSUE #306: Add support for pre-uninstall scripts
If I understand it correctly, the package kit now supports pre-uninstall scripts (see https://dev.haiku-os.org/ticket/13427). It seems that HaikuPorter doesn't support them yet.

jmairboeck:
This is analoguous to post-install scripts.

Defines a new recipe attribute PRE_UNINSTALL_SCRIPTS and new directory variables $preUninstallDir and $relativePreUninstallDir.

Fixes #306.

### DIFF
--- a/HaikuPorter/Package.py
+++ b/HaikuPorter/Package.py
@@ -400,6 +400,9 @@ class Package(object):
 				self._writePackageInfoListQuotePaths(infoFile,
 					self.recipeKeys['POST_INSTALL_SCRIPTS'],
 					'post-install-scripts')
+				self._writePackageInfoListQuotePaths(infoFile,
+					self.recipeKeys['PRE_UNINSTALL_SCRIPTS'],
+					'pre-uninstall-scripts')
 
 			# Generate SourceURL lines for all ports, regardless of license.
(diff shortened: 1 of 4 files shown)


jmairboeck:
I added a little drive-by fix because PROVIDES and POST_INSTALL_SCRIPTS were sorted incorrectly in RecipeAttributes.

waddlesplash (approved):

jmairboeck:
What is needed to get this deployed? Or do the builders update haikuporter from git automatically?

waddlesplash:
It has to be done manually. I just logged into the builders and did that.

jmairboeck:
I finally found the cause of the spurious "PRE_UNINSTALL_SCRIPTS" fatal warning at the end that appears for old recipes (`--debug` option has helped): This is a KeyError that occurs in the policy check for this, because it may not be in the cached version of the recipe.

The solution/workaround is to delete `port.recipe` from the workdir to force reparsing so that the new attribute is included in the known recipe attributes.

(at least @Begasus also had this a few times I think after this was introduced.)

PR #311, opened 2025-01-02
Add missing qoutes for meson scriplet

Begasus:


Begasus:
Without these double quotes I wasn't able to launch a meson build. First run was correct (debug enabled):
````
error: invoking meson with --buildtype=debugoptimized without debug info packages specified
````
Then changed buildtype to release and got:
````
error: invoking meson with --buildtype=release with debug info packages specified
````

File: HaikuPorter/ShellScriptlets.py [hunk 1]
```diff
@@ -353,10 +353,10 @@ def getShellVariableSetters(shellVariables):
 		echo "error: invoking meson without --buildtype argument"
 		echo "note: you probably want --buildtype=release or --buildtype=debugoptimized"
 		exit 1
-	elif [[ "$*" = *buildtype=release* ]] && [ -n $DEBUG_INFO_PACKAGES ]; then
+	elif [[ "$*" = *buildtype=release* ]] && [ -n "$DEBUG_INFO_PACKAGES" ]; then
```
jmairboeck, line 356: Why do these mix `[[` and `[`?

jmairboeck (approved):

HaikuPorter/ShellScriptlets.py
Begasus: Beats me, didn't spot it, will run some checks and update if needed

Begasus: Looking at the cmake pattern they match, @korli should know. :)

korli: Indeed [[ would be preferred.

Begasus: @korli in the cmake part there are also only one [ ] at the end of the lines, so I didn't change them here, should those also be double for the last statement?
````
	if [[ "$*" != *--install* ]] && [[ "$*" != *--build* ]]; then
		if [[ "$*" != *CMAKE_BUILD_TYPE* ]]; then
			echo "error: invoking cmake without CMAKE_BUILD_TYPE specified!"
			echo "note: you probably want -DCMAKE_BUILD_TYPE=Release or -DCMAKE_BUILD_TYPE=RelWithDebInfo"
			exit 1
		elif [[ "$*" = *CMAKE_BUILD_TYPE=Release* ]] && [ -n "$DEBUG_INFO_PACKAGES" ]; then
			echo "error: invoking cmake with -DCMAKE_BUILD_TYPE=Release with debug info packages specified"
			exit 1
		elif [[ "$*" = *CMAKE_BUILD_TYPE=RelWithDebInfo* ]] && [ -z "$DEBUG_INFO_PACKAGES" ]; then
			echo -n "error: invoking cmake with -DCMAKE_BUILD_TYPE=RelWithDebInfo without debug info "
			echo "packages specified"
			exit 1
		fi
	fi
````

Begasus: I'll leave the brackets for the next one cleaning this up. At least this fixes building with meson.

PR #312, opened 2025-01-07
Sort the output of --list-dependencies.

OscarL:
Had this small patch lying around for too long.

Makes reading/comparing the output `--list-dependencies` easier.

### DIFF
--- a/HaikuPorter/Main.py
+++ b/HaikuPorter/Main.py
@@ -513,17 +513,21 @@ class Main(object):
 		print('')
 
 		print('packages that need to be built:')
+		output = []
 		for dependency in buildDependencies:
 			packageInfoFileName = os.path.basename(dependency)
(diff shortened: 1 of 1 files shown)


Begasus:
LGTM, list is sorted which makes it indeed easier to read. I'm just confused (but I think I already mentioned that a while back at IRC) about the actual list produced. :)

Begasus:
Like: I'm seeing `uchardet_x86` when I run this on `hp tiff --list-dependencies`, I don't get why or how uchardet would be a dependency for tiff, for other packages maybe, but still ... ends up in a pit hole.

Begasus:
> AFAICS... `--list-depedencies` lists all the direct and indirect build-requirements, so you get this long beast.

I get that, but fwiw, I never found it something useful to check out, underlying dependencies should be shown there, my 2 cents.

Anyway, for it's purpose this looks fine to me. 👍

PR #313, opened 2025-01-07
Don't attempt patchset extraction if work dir is not a git repo.

OscarL:
Might avoid data loss (due to an empty .patchset overwritting an existing one).

Don't ask me how I know :-(

### DIFF
--- a/HaikuPorter/Source.py
+++ b/HaikuPorter/Source.py
@@ -405,6 +405,10 @@ class Source(object):
 			sysExit("Can't extract patchset for " + self.sourceDir
 					+ u" as the source directory doesn't exist yet")
 
+		if not self._isInGitWorkingDirectory(self.sourceDir):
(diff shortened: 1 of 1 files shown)


Begasus:
Happened to me too ... 🤣 
+1 for me, but needs to be checked first. :)

HaikuPorter/Source.py
```diff
@@ -405,6 +405,10 @@ def extractPatchset(self, patchSetFilePath, archPatchSetFilePath):
 			sysExit("Can't extract patchset for " + self.sourceDir
 					+ u" as the source directory doesn't exist yet")
 
+		if not os.path.exists(self.sourceDir + '/.git'):
```
korli, line 408: This should use if not self._isInGitWorkingDirectory(self.sourceDir) instead.

waddlesplash (approved):

PR #317, opened 2025-01-26
policy: add pkgconfig rule

X547:
pkgconfig entries in develop/lib[<arch>]/pkgconfig must be declared as pc:*[_<arch>].


jmairboeck:
Aren't these (mostly) redundant with the existing `devel:` entries?

X547:
No, `devel:` entries often do not match pkgconfig names. Single devel package may also provide multiple pkgconfig entries.

It is a pain to search package names that match build requirements.

pulkomandy:
26 janvier 2025 à 12:01 "X512" ***@***.*** ***@***.***?to=%22X512%22%20%3Cnotifications%40github.com%3E > a écrit:

> 
> No, devel: entries often do not match pkgconfig names. Single devel package may also provide multiple pkgconfig entries.
> 
> It is a pain to search package names that match build requirements.
> 
Can we fix devel: entries to match pkgconfig names? Or do they have to match something else? (they currently match library names, but I'm not sure they have to, or we can have two devel entries if the two disagree in a few cases).

One problem here is that every provide of every package is also copied into the main repository file (so that "pkgman install xxx" can find them there and work offline), so, adding more provides makes the package repo much larger.

So, if we can avoid having 90% of devel packages provide both a pc: and devel: entry with the same name, it would be nice to do it that way.

X547:
As can be seen in code above changed part, `devel:` matches library name in `develop/lib` directory. But pkgconfig names don't necessary match library names. It can be header-only library that have no any libraries in `develop/lib` directory. It may be pkgconfig entry that refer multiple libraries. pkgconfig names are arbitrary and may be not consistent (may have or not have "lib" prefix etc.).

> so, adding more provides makes the package repo much larger.

By this logic we should also avoid long package descriptions or porting too much software, that if bad idea. Also without actually trying, we can't know how larger repo file will become, maybe not so much.

It may be good idea to implement some kind of delta updates of repo file instead of fully downloading it on each run. Most of packages stay the same after repo update.

pulkomandy:
26 janvier 2025 à 15:54 "X512" ***@***.*** ***@***.***?to=%22X512%22%20%3Cnotifications%40github.com%3E > a écrit:


> 
> As can be seen in code above changed part, devel: matches library name in develop/lib directory. But pkgconfig names don't necessary match library names. It can be header-only library that have no any libraries in develop/lib directory. It may be pkgconfig entry that refer multiple libraries. pkgconfig names are arbitrary and may be not consistent (may have or not have "lib" prefix etc.).
> 
I know about the pkg-config situation. My question is more:

- Why do "devel:" entry have to match library names? Is that useful? Would it be a problem if we changed it to match pkgconfig files (when available)? I guess it is useful when your compiler says "-lxxx: library not found" and you want to install the matching package
- Second question, are pkgconfig files often identical with library names? If so, maybe we can use devel: entries for both. If most packages have a pkgconfig entry matching the library name, they get a single "devel:" entry also matching that. The few packages that have mismatches get two "devel:" entries. That is saving some space and processing time(maybe that saving is not large enough to be relevant) if a lot of libraries and pkgconfig files are matched. Otherwise, it's a bad idea, and introducing a separate namespace is better because it's cleaner and more explicit.


> 
> > 
> > so, adding more provides makes the package repo much larger.
> > 
> 
> By this logic we should also avoid long package descriptions or porting too much software, that if bad idea. Also without actually trying, we can't know how larger repo file will become, maybe not so much.
> 
I'm just asking if we can reduce the problem in this one specific situation. If the answer is no, that is fine. but if the answer is yes, there is no reason to waste resources.


> 
> It may be good idea to implement some kind of delta updates of repo file instead of fully downloading it on each run. Most of packages stay the same after repo update.
> 
Yes, there are other ways to improve this, of course.

waddlesplash:
I would be interested in examples of this.

IMO we should not add more provides just for pkg-config files, at least in general (maybe there are specific cases where it makes sense). If the pkg-config file is just some other abbreviated form of the name then let's just make sure that abbreviated form appears in the package summary/description so `pkgman search` picks it up. Otherwise, maybe what we want/need is a "files database" that collects all files from all packages in some central place and lists them, separately from the repo file of course as we don't want to make it any larger.

X547:
> I would be interested in examples of this.

Use case would be something like:

1. `meson introspect --scan-dependencies meson.build`
2. `pkgman install pc:dep1 pc:dep2 pc:dep3 pc:dep4 pc:dep5`
3. `mkdir build; cd build; meson setup ..; ninja`

X547:
> - Why do "devel:" entry have to match library names? Is that useful?

`devel:` entries are actually not very useful for projects that use pkgconfig. Just specifying library names are often not enough, additional include paths and defines are often needed to compile program that use library.

It will be better to specify `pc:` entries in recipe build requires instead of `devel:` entries because this is how build process is actually works.

waddlesplash:
No, I meant, what actually existing software has such .pc files that don't match either the software (package) name, or the library name?

waddlesplash:
The only one of those where the package name doesn't correspond to either the pkg-config name or library name is glvnd, and that does provide a `libGL`. So I don't see where the disconnect is here. All of those would be obvious.

pulkomandy:
x512 explained very clearly why this is needed:

- Header only libraries may have only a pkg-config file and no libraries
- Library name may not match pkg-config name (this is the case for example for Lua)

"freetext" data in the description won't do it. First of all, it's clumsy (forcing people to write some specific words in the description, while we have metadata for it). Secondly, it's harder to parse and so would prevent making scripts or tools that automate things. For example, think of how Debugger can autoinstall needed debuginfo package. Surely an IDE may want to do the same, by scanning a buildsystem file or buildtool output, and automatically offering to install the needed development packages. For this to work, there has to be a predictable name related to what the buildsystem is looking for. This can be either a library name (if looking directly at -lxxx parameters in the linker command line), or a pkgconfig filename.

For the same reason, storing that outside the repository file makes little sense: providing that information is exactly the job of the repository file.

Another idea would be to rename the pkgconfig files. Unfortunately, doing so defeats their purpose, which is to have a predictable name to look for, no matter what platform you're in. So it's important that upstream pkgconfig names are kept.

So the remaining options are:

- Separate devel: and pc: namespaces, cleaner but potentially a lot of duplication,
- Using the same devel: namespace, less explicit, opens room for conflict if a pkgconfig file in one package has the same name as a library in another package, but avoids having two entries when the lib and pkgconfig file have the same name.

jmairboeck:
The third option would be to not require a `devel:` entry if there is a `pc:` entry which covers it. But that would maybe be too much of a breaking change.

waddlesplash:
> Header only libraries may have only a pkg-config file and no libraries

In that case I would expect a `devel:` entry for the pkg-config file.

> Library name may not match pkg-config name (this is the case for example for Lua)

In that case (but only in that case) it may make sense to have a separate provide. So, the policy check would then be that there is *some* provide matching the pkg-config name, or the pkg-config name with `lib` on the front.

But if there are only a very small number of applications that fit that paradigm, like Lua, then a full policy check may not make sense and we should just add the provides for the particular packages as special cases.

X547:
> In that case (but only in that case) it may make sense to have a separate provide

How it will work? Add some expressions support to pkgman like `pkgman install (devel:dep1 || pc:dep1) (devel:dep2 || pc:dep2) (devel:dep3 || pc:dep3)`? Without it, it will be not possible to automatically install pkgconfig dependencies.

waddlesplash:
`pkgman search` should be used regardless because multiple packages may provide the same library already (e.g. ICU 74 vs 75) and the user should be offered a choice of what to install (or not to install), one shouldn't just blindly run `pkgman install` in most cases.

X547:
> multiple packages may provide the same library already (e.g. ICU 74 vs 75) 

Dependency solver support specifying resolvable versions.

X547:
Found pkgconfig files installed to wrong location while working on this: https://github.com/haikuports/haikuports/issues/11694.

PR #321, opened 2025-03-23
fix "any" arch subpackages on secondary arch

### LINKED ISSUE #320: ARCHITECTURES_foo="any" doesn't work for secondary arch
Subpackages with "any" architecture don't actually end up being "any" when using a secondary architecture, but they instead get the same primary architecture as the main package (i.e. "x86_gcc2" in practice).

See for example, the lilypond_x86_doc packages on x86_gcc2: https://depot.haiku-os.org/#!/pkg/lilypond_x86_doc/haikuports/haikuports_x86_gcc2/2/24/4/-/1/x86_gcc2?bcguid=bc618-OSBT

Besides, I think, they should drop the secondaryArchSuffix from the resulting package name in this case (e.g. being called "lilypond_doc" instead of "lilypond_x86_doc" in the above example) so that the result is the same as on a primary architecture. This will also be needed to eventually deduplicate "any" packages in the repositories (see #234).

jmairboeck:
When determining the architecture of a package, check whether it is "any" before checking the secondary architecture.

To make "any" packages identical to the case when not using a secondary architecture, don't include the secondaryArchSuffix in the package name in this case. This is a breaking change! Any packages using this must update the PROVIDES of the subpackage and should add a REPLACES entry to facilitate upgrades.

Add support for this change to ShellScriptlets getPackagePrefix, which is also used by packageEntries.

Fixes #320.

### DIFF
--- a/HaikuPorter/Package.py
+++ b/HaikuPorter/Package.py
@@ -96,6 +96,8 @@ class Package(object):
 				self.architecture = port.targetArchitecture
 			else:
 				self.architecture = Architectures.SOURCE
+		elif Architectures.ANY in self.recipeKeys['ARCHITECTURES']:
+			self.architecture = Architectures.ANY
 		elif ((port.secondaryArchitecture is not None and
 			  port.secondaryArchitecture in self.recipeKeys['SECONDARY_ARCHITECTURES']) or
 			  port.targetArchitecture in self.recipeKeys['ARCHITECTURES']):
@@ -104,8 +106,6 @@ class Package(object):
 			# is the same as the target architecture, except for "_cross_"
 			# packages, which are built for the host on which the build runs.)
 			self.architecture = port.hostArchitecture
-		elif Architectures.ANY in self.recipeKeys['ARCHITECTURES']:
-			self.architecture = Architectures.ANY
 		else:
 			sysExit('package %s cannot be built for architecture %s'
 					% (self.versionedName, port.targetArchitecture))
@@ -142,8 +142,9 @@ class Package(object):
 
 	def getStatusOnSecondaryArchitecture(self, architecture,
 			secondaryArchitecture):
-		# check the secondary architecture
-		if secondaryArchitecture:
+		# check the secondary architecture (but ANY takes precedence)
+		if (secondaryArchitecture
+			and not Architectures.ANY in self.recipeKeys['ARCHITECTURES']):
 			secondaryStatus = Status.UNSUPPORTED
 			secondaryArchitectures = self.recipeKeys['SECONDARY_ARCHITECTURES']
 			if secondaryArchitecture in secondaryArchitectures:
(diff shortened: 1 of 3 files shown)


jmairboeck:
I tested this with perl_x86, which now creates a perl_doc-any package instead of perl_x86_doc-x86_gcc2 on 32 bit.

HaikuPorter/ShellScriptlets.py
```diff
@@ -94,8 +94,14 @@ def getShellVariableSetters(shellVariables):
 		local linksDir="$packageLinksDir/$packageName-$portFullVersion"
 		local packagePrefix="$linksDir/.self"
 		if [ ! -e "$packagePrefix" ]; then
-			echo >&2 "packageEntries: warning: \"$packageSuffix\" doesn't seem to be a valid package suffix."
-			exit 1
+			# try again with the base name
+			local packageName="${portName/%$secondaryArchSuffix}_$packageSuffix"
```
jmairboeck, line 98: Should we introduce a `portBaseName` variable without the secondaryArchSuffix?

waddlesplash: That probably makes sense?

jmairboeck: done

waddlesplash:
So, I'm not sure this can be deployed to the builders without updating the buildmaster itself, since the buildmaster parses the recipe itself in order to find out what the package names will be, and this changes that behavior. CC @mmlr.

(We are currently blocked on upgrading the buildmaster on the S3 migration...)

jmairboeck:
Ah, yes, the buildmaster web interface shows the "resulting packages" of a pending build run, so it has to parse the recipes.

jmairboeck:
I think I found an issue here: https://github.com/haikuports/haikuporter/blob/6cc8967e179d693f942fd27e628eb332ca4eb979/HaikuPorter/BuildMaster.py#L301-L308

This will be a problem if we have ports that are buildable on both primary and secondary architectures, because now the package names will clash. Do we have any of these?

waddlesplash:
Yes, we have a lot of library packages that fit that bill.

HaikuPorter/Package.py
```diff
@@ -142,8 +142,9 @@ def isBuildableOnArchitecture(self, architecture):
 
 	def getStatusOnSecondaryArchitecture(self, architecture,
 			secondaryArchitecture):
-		# check the secondary architecture
-		if secondaryArchitecture:
+		# check the secondary architecture (but ANY takes precedence)
+		if (secondaryArchitecture
+			and not Architectures.ANY in self.recipeKeys['ARCHITECTURES']):
```
jmairboeck, line 147: I'm not sure if this change was actually needed, I'll check this again. I think this is what causes "any" packages now being tried to be built twice in buildmaster mode (see e.g. https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=8186&viewMode=expanded).

PR #322, opened 2025-03-23
ShellScriptlets: better support for PACKAGE_NAME and PACKAGE_VERSION in getPackagePrefix

jmairboeck:
and therefore packageEntries.

Untested because nothing seems to use this currently.

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -76,34 +76,23 @@ set -e
 
 getPackagePrefix()
 {
-	# Usage: getPackagePrefix <packageSuffix>|<packageFullVersionedName>
+	# Usage: getPackagePrefix <packageSuffix>
 
+	local packageSuffix="$1"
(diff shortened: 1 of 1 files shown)


jmairboeck:
It looks like the "packageFullVersionedName" variant of packageEntries and getPackagePrefix was just introduced to support PACKAGE_NAME in https://github.com/haikuports/haikuporter/commit/463f228e9554935be64f3bfdab5a269ae2a7a93c. Do we even need that variant then?

jmairboeck:
The "full" form of `packageEntries` and `getPackagePrefix` doesn't seem to be used in haikuports (according to `git grep`), so I think we can remove it then.

jmairboeck:
I removed the "packageFullVersionedName" variant now (in a separate commit).

Good to merge?

jmairboeck:
Both cases work (tested with pip and setuptools respectively). Given that this isn't actually used yet, I'm going to merge this.

@waddlesplash: I think you can wait with deploying this until #321 is also merged (if it is going to be).

PR #329, opened 2025-06-08
Only reset to HEAD if the PATCH function fails, not to ORIGIN

### LINKED ISSUE #284: When the PATCH function fails, haikuporter resets to ORIGIN, removing patchset commits
I noticed this when updating `texlive_core`, which uses both a patchset and a PATCH function. The PATCH function failed (because it needs autoconf 2.72 apparently) and HaikuPorter resetted the internal git repo back to ORIGIN, effectively discarding the patchset commits. The `flag.patchset` file was not deleted, however, so that results in an inconsistent state.

Do we need a new tag that is applied after the patchset and reset to that instead if it is present, or could we just reset to HEAD instead when the PATCH function fails? If the resetting to ORIGIN is retained, at least the `flag.patchset` should be removed in this case.

jmairboeck:
And also do it if there are no file-based patches. These and the PATCH function are orthogonal features.

Fixes #284.


HaikuPorter/Source.py
```diff
@@ -366,12 +366,14 @@ def patch(self, port):
 		if patched:
 			port.setFlag('patchset', self.index)
 
-		return patched
-
-	def reset(self):
+	def reset(self, resetToOrigin=True):
```
jmairboeck, line 369: Should I maybe change the parameter to a string defaulting to 'ORIGIN' and call it with 'HEAD' in the patch function case?

PR #333, opened 2025-07-07
Port: don't use requires initially

korli:
DependencyResolver now makes a copy of requireTypes so that it can be modified.

Should help with https://github.com/haikuports/haikuports/issues/12417


waddlesplash:
CC @kallisti5

kallisti5:
This change makes sense to me.  I'm not deeply knowledgeable about  the internals of package dependency solving in  haikuporter however (nor a Python expert).

Maybe @mmlr should review this one?

HaikuPorter/DependencyResolver.py
```diff
@@ -63,7 +63,7 @@ class DependencyResolver(object):
 	def __init__(self, buildPlatform, requiresTypes, repositories, **kwargs):
 		self._providesManager = ProvidesManager()
 		self._platform = buildPlatform
-		self._requiresTypes = requiresTypes
+		self._requiresTypes = list(requiresTypes)
```
pulkomandy, line 66: I'd say `requiresTypes.copy()` (or even `.deepcopy()` if needed) would make the intent clearer if it's just about making a copy? Or is the cast to a list also useful for another reason?

pulkomandy: yes, it's a constructor that will make a list from whatever you pass as a parameter.

If the parameter is already a list, it makes a copy of it. If it's something else, it (attempts to) convert it into a list.

Using the copy() method would preserve the original type of the object.

kallisti5:
Looks good to me.  With @pulkomandy 's concerns addressed... I think we're good! 

Thanks @korli :-)

Issue #192, opened 2019-10-10
Python base packages don't provide anything ...


Begasus:
I've seen it passing by while working on some of the different python_version packages, python versions are defined and they build ok.
You get 4 different packatges, 1 base (empty), and one for each python version (2.7, 3.6 and 3.7)
Is there a way to not build the base package or set it to 2.7 as a base and 3.6 and 3.7 as second and third package?

pulkomandy:
Why would 2.7 be the base? Python 2.7 is deprecated and should not be the default. If you want to make one of these the base, use 3.7. But then we're in for a lot of renaming when 3.8 comes in.

Removing the base package would be nice, we should add that to Haikuporter if not already possible.

Begasus:
2.7 was the first that came to mind here :)

korli:
Proposal: DISABLE_MAIN_PACKAGE="yes"

alaviss:
An another way is to make them a meta package, so people installing them will get all `_python*` variants.

Begasus:
Both valid options, but we need one way to go with this :)

Glutexo:
Sorry for the dumb question. How to install Python 3 when Haikuporter requires Python 3 for running? With default Haiku Beta 2, only Python 2.7 is available and running _haikuporter_ results in

```
env: 'python3': No such file or directory
```

pulkomandy:
`pkgman install cmd:python3` should do it

Glutexo:
You are right! Thank you for your help! 🙇🏻‍♂️ Hopefully this helps anybody else with the same problem.

Begasus:
Some "main" packages are not empty (for instance pytest, it provides the binary in the main package, for this one I also saw that it uses python 3.7 modules, so maybe that one should be dimmed to be only build with 3.7 (pushed a version earlier, but my skills on python are not that great)) :)

jmairboeck:
This would probably also solve the name conflict between [python pathtools](https://github.com/haikuports/haikuports/blob/master/dev-python/pathtools/pathtools-0.1.2.recipe) and [perl pathtools](https://github.com/haikuports/haikuports/blob/master/dev-perl/pathtools/pathtools-3.75.recipe) nicely when the python one doesn't build a main package any more.

PR #335, opened 2025-07-19
fix "any" arch ports being considered buildable on secondaryArch

jmairboeck:
This fixes a regression from e1d6715aea62d5b5c0e00401a863ee197eec2258, which erroneously considered 'any' arch ports (where the main package is "any") as buildable on secondary architectures. Besides that this doesn't make sense logically, in practice these packages are never buildable because the value of $prefix is wrong in this case because the name of the main package doesn't match the port name (and $prefix is used by all recipes directly or indirectly because it is the target for all files to be installed to). The above commit only added support for getPackagePrefix for the subpackages, not the main one.

Additionally, this bug caused the package-to-port associations in the repository to become wrong and always point to the (last) secondaryArch port instead of the primary one because secondary architectures are processed after the primary architecture.

This change fixes the bug by marking ports with an "any" arch main package as explicitly unsupported. This can't be done at the package level, because it doesn't know whether it is the main package or a subpackage.


jmairboeck:
This disallows the main package being "any" having some architecture-specific subpackages (with secondaryArch), but I don't think we have such a case right now anyway. To support that, we would need `DISABLE_MAIN_PACKAGE` (see #192) to avoid the conflict on the duplicated main packages and fix the above mentioned issue with `$prefix` so that creating such a recipe wouldn't be so tedious.

Begasus:
Checks with themes and kitinerary worked out fine, so +1 for me.

jmairboeck:
> This disallows the main package being "any" having some architecture-specific subpackages (with secondaryArch), but I don't think we have such a case right now anyway. To support that, we would need `DISABLE_MAIN_PACKAGE` (see #192) to avoid the conflict on the duplicated main packages and fix the above mentioned issue with `$prefix` so that creating such a recipe wouldn't be so tedious.

Just to document this, what I meant what doesn't work now is something like this:

Example 1: architecture-specific subpackage which is buildable everywhere:

```sh
ARCHITECTURES="any"
if [ -n "$secondaryArchSuffix" ]; then
	# this is needed to not build the main package multiple times
	DISABLE_MAIN_PACKAGE="yes" # or however this is done then
fi

ARCHITECTURES_binaries="all"
SECONDARY_ARCHITECTURES_binaries="x86"
```

Example 2: architecture-specific subpackage not for gcc2:

```sh
if [ "$effectiveTargetArchitecture" = "x86_gcc2" ]; then
	# this is needed because the main package determines the state of the whole port
	ARCHITECTURES="" # or maybe "!any"
else
	ARCHITECTURES="any"
fi

ARCHITECTURES_binaries="all !x86_gcc2"
SECONDARY_ARCHITECTURES_binaries="x86"
```

PR #337, opened 2025-08-03
don't include secondaryArchSuffix in docDir

jmairboeck:
Documentation is usually architecture-independent, and there are multiple packages in haikuports which in fact use an "any" arch subpackage for the documentation. These packages then end up being different if the `$secondaryArchSuffix` is included in the documentation directory name.

The alternative would be that these recipes can't use `$docDir`, and by extension `runConfigure` and `$cmakeDirArgs` without overriding the doc dir, if they install anything in it.

After this change, packages that are built for both primary and secondary architectures should only include the documentation in only one of them to avoid file conflicts (although I think most of them already do, and on the other hand, these conflicts are probably not a problem if the contents are the same).

----

RFC: do you think this is a good idea?


waddlesplash:
I don't think this is a good idea. For libraries it's the wrong thing to do for sure, since as you note they may be different between architectures.

jmairboeck:
Ok, I'm closing this. This must be fixed in the affected recipes instead (i.e. most of the ones in this list: https://github.com/haikuports/haikuports/issues/12108).

PR #344, opened 2025-08-30
runConfigure: extend CFLAGS checks to check debug flags

jmairboeck:
Similar to what the cmake and meson functions already check.

----

I hope there aren't any non-debug related compiler flags starting with "-g". Do we need a way to opt out of this maybe? (like with `unset cmake` in the cmake case, which isn't so easy in this case)


waddlesplash:
Opting out is easy, just don't call `runConfigure` and invoke it manually.

PR #347, opened 2025-10-15
Adds a config option to set the packages compression level.

OscarL:
Both the command line and config file can be used to set the value.

### DIFF
--- a/HaikuPorter/Configuration.py
+++ b/HaikuPorter/Configuration.py
@@ -12,7 +12,7 @@ import re
 from .ConfigParser import ConfigParser
 from .Options import getOption
 from .RecipeTypes import Extendable, MachineArchitecture, YesNo
-from .Utils import sysExit
+from .Utils import sysExit, warn
 
 
 def which(program):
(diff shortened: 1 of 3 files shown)


HaikuPorter/Configuration.py
```diff
@@ -146,6 +146,15 @@ def isExecutable(path):
...
 	},
+	'PACKAGE_COMPRESSION_LEVEL': {
+		'type': bytes,
+		'required': False,
+		'default': None,
+		'extendable': Extendable.NO,
+		'indexable': False,
+		'optionAttribute': 'compressionLevel',
+		'setAttribute': 'packageCompressionLevel',
+	},
```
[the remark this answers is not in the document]
OscarL: Mmm not sure. My intention here is to be able to lower (or disable) the compression while working on recipes (as that steps can take quite a while otherwise).

Basically doing what `HAIKU_PACKAGE_COMPRESSION_LEVEL` does for building Haiku locally.

Do you have some particular use-cases in mind? Couldn't those be handled locally with a custom .sh that does `hp foobar --package-compression-level x`?

Or you envision use cases where such a setting would be needed to work also on the builders?

waddlesplash (approved):

HaikuPorter/Configuration.py
```diff
@@ -146,6 +146,15 @@ def isExecutable(path):
 		'optionAttribute': 'commandPackage',
 		'setAttribute': 'packageCommand',
 	},
+	'PACKAGE_COMPRESSION_LEVEL': {
+		'type': bytes,
```
pulkomandy: why not set it to type int instead?

OscarL: Because I'm dumb :-) (did a quick look for other types used, didn't saw int... went with what was already there).

Thanks for the note, will try it today. (same for the comment on Options.py below).

PR #349, opened 2025-11-09
defineDebugInfoPackage: change first argument to basePackageSuffix

### LINKED ISSUE #348: defining multiple debuginfo packages doesn't work
Although the [implementation of `defineDebugInfoPackage`](https://github.com/haikuports/haikuporter/blob/8f8ce95df466192d7dbca9bcde5785f16455c806/HaikuPorter/ShellScriptlets.py#L107-L157) looks like it should support it almost, in practice it doesn't because the `packageSuffix` it uses is hardcoded to just "debuginfo", which causes a conflict when trying to invoke it multiple times.

Multiple debuginfo packages could be useful for example for recipes which define a separate "tools" package to also separate the debuginfos accordingly.

Additionally, as it is now, the first parameter of `defineDebugInfoPackage` (basePackageName) is a bit redundant because `$portName` could just be used for that. The printed usage error message says that it should be the packageSuffix instead (but not the comment and the actual implementation). Should we maybe change that so that it is consistent and multiple debuginfo packages can be defined?

**Proposal:** change the first parameter to the package suffix without the "debuginfo" part, i.e. the default would be an empty argument (""). This is a breaking change, but recipes could be fixed rather easily by a mass replacement. The existing basePackageName could be replaced by `$portName` if the suffix is empty or `${portName}_$packageSuffix` otherwise (or `$PACKAGE_NAME` if that is defined).

```sh
if [ -z "$1" ]; then
	local basePackageName=$portName
	local packageSuffix=debuginfo
else
	local basePackageName=${portName}_$1 # TODO: support PACKAGE_NAME if defined
	local packageSuffix=${1}_debuginfo
fi
local packageName=${basePackageName}_debuginfo # TODO: support PACKAGE_NAME if defined
```

What do you think about this?

jmairboeck:
This allows defining multiple debuginfo packages by not hard-coding the used packageSuffix for the debuginfo package any more but deriving it from the basePackageSuffix instead.

**Note: This is a breaking change!** All invocations of defineDebugInfoPackage in haikuports must be changed. The default case to define the main debuginfo package is to use an empty string (i.e. "") as the first argument now. To define a debuginfo package for a subpackage, use its package suffix instead (similar to packageEntries and getPackagePrefix).

Also adds support for PACKAGE_NAME and PACKAGE_VERSION overrides, both for the referenced base package and the debuginfo package itself. In both cases, they must be declared before defineDebugInfoPackage is called.

Fixes #348.

----

Question: should the `debugInfoSuffix` also be changed for subpackages? (i.e. the base package name and version that is added to debug info file names)

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -110,7 +110,6 @@ defineDebugInfoPackage()
 	fi
 
 	local destDir=$debugInfoDir
-	local debugInfoSuffix="($portRevisionedName)"
 	if [ "$1" = "--directory" ]; then
 		destDir="$2"
 		shift 2
@@ -122,6 +121,7 @@ defineDebugInfoPackage()
 	if [ -z "$basePackageSuffix" ]; then
 		local basePackageName=$portName
 		local basePackageVersion=$portVersion
+		local debugInfoSuffix="($portRevisionedName)"
 		local packageSuffix=debuginfo
 	else
 		eval "local basePackageName=\"\$PACKAGE_NAME_$basePackageSuffix\""
@@ -132,6 +132,7 @@ defineDebugInfoPackage()
 		if [ -z "$basePackageVersion" ]; then
 			local basePackageVersion=$portVersion
 		fi
+		local debugInfoSuffix="($basePackageName-$basePackageVersion-$REVISION)"
 		local packageSuffix=${basePackageSuffix}_debuginfo
 	fi
 	eval "local packageName=\"\$PACKAGE_NAME_$packageSuffix\""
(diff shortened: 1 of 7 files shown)


jmairboeck:
> Another question: should this handle inherited attributes (like SUMMARY, DESCRIPTION etc.) to automatically "inherit" them from the (non-default) base package, if they have been overridden for that?
> 
> (I guess this can get pretty complex if we really want to go that far ...)

I think this is easier to be handled in the recipes. Let's say: if you override these variables for the debuginfo base package, you also have to do it for the debuginfo package manually for consistency.

I'll add that to the glib recipe in the example PR.

PR #350, opened 2025-11-24
buildmaster: Limit rescheduling of transient errors

kallisti5:


kallisti5:
10 might be too low... (the transient errors are supposed to be re-attempted because they could be network disconnects, etc)  However, infinite attempts is too many. (especially when every build create a log file)

kallisti5:
> There is a problem with this when there are multiple builders. In case builders have issues and rack up these transient errors, while there still is at least one working builder, they will cause builds to be marked as failed just because the working builder is busy and doesn't get around to it. This happens regularly in the riscv64 case. Take buildrun 622:

I don't follow on this. The scheduled builds loop with transient errors when a builder isn't available?  That's counter to what I see in the code at the following locations:

* https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/BuildMaster.py#L405 picking a builder for a scheduledJob
* https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/BuildMaster.py#L422 starting a thread building that job on the assigned builder.
* https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/BuildMaster.py#L477 appending the build number to the scheduled job on that builder
* https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/BuildMaster.py#L491 failure, scheduled Build goes back into the pool of scheduled builds (that can indeed run on any builder).   Gets re-assigned in L405 looping back to the beginning step.

Even if the build gets scheduled on the same (or alternative builder), it seems logical that "10 failed attempts to build a single port is enough"

I could easily be misunderstanding something though with the above.. the code for this scheduling is pretty complected.


> Which presumably comes from here, as this is the only open_session I can find: https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/Builders/RemoteBuilderSSH.py#L419
>
> So this implies that the transport is None and it should really be treated as a connection error and cause the builder to be marked lost/reconnecting. Other places check if the transport is alive, but in this case that is never validated before trying to use the transport.
>
> The _sftpClient() just above _runCommand() looks broken as well, as it will just return None and none of its users check for that...

Yup, 100% possible  there are a few edge cases triggering this in the new ssh code I put in place a few months ago.    We have also seen this looping failure when the date of a builder is off (resulting in a continuous loop of git pull failures, like in 622.

PR #352, opened 2025-11-26
meson: Exclude submodule from triggering the --buildtype check

nilFinx:
Required for a new port that I'm going to release soon (I can just use /bin/meson to avoid it, but I figured that I should push this fix)

### DIFF
--- a/HaikuPorter/ShellScriptlets.py
+++ b/HaikuPorter/ShellScriptlets.py
@@ -384,7 +384,8 @@ cmake()
 meson()
 {
 	if [[ "$*" != *buildtype* ]] && [[ "$*" != compile* ]] \
-			&& [[ "$*" != install* ]] && [[ "$*" != test* ]]; then
+			&& [[ "$*" != install* ]] && [[ "$*" != test* ]] \
+			&& [[ "$*" != submodule* ]]; then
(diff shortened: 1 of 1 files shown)


waddlesplash:
What's this command for? We don't want to download code inside the chroot. (In fact it may be a good idea to prevent network access inside the chroot entirely, maybe unless some special flag is set on the recipe, but we haven't gone that far yet...)

waddlesplash:
We require all downloads of source code to be done by HaikuPorter and verified by hashes specified in the recipe itself. So, this command shouldn't be used at HaikuPorts then.

waddlesplash:
> then projects that depends on many submodules will be more time consuming for prett much no reason

Most projects with many submodules often provide release tarballs with all the submodules included. Or, if the submodules are third-party dependencies, then those dependencies should each get their own recipes. Or, in the case of things like Rust, we have scripts in this repository which will take a Cargo file and automatically create a recipe with everything specified.

> and I'm pretty sure that mamy of them uses it to download something.

Most of them likely use it to fetch commit hashes or do other such things. If they're using it to download, they probably shouldn't be.

> airgapped recipes isn't a good idea.

Why not?

PR #354, opened 2025-12-03
Add tools to valid suffixes

Begasus:
This silenced haikuporter report for it not being a valid suffix

### DIFF
--- a/HaikuPorter/Package.py
+++ b/HaikuPorter/Package.py
@@ -38,6 +38,7 @@ class PackageType(str):
 	DOCUMENTATION = 'doc'
 	GENERAL = 'general'
 	SOURCE = 'source'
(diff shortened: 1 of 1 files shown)


Begasus:
Adding one for the likes of "qt5" wouldn't make sense imho, for tools I think it should, as imho it's a more valid suffix used.
If the predefined summary suffix for tools fixes this, I'm all in. :+1:

jmairboeck:
While haikuporter updated the dependency infos for the mass change in https://github.com/haikuports/haikuports/pull/13694, I also noticed one "qt6" suffix and a few "bin" suffixes show up with that error message.

jmairboeck:
I'll have look if I can find the issue.

PR #357, opened 2025-12-31
make Architecture and MachineArchitecture values case-sensitive

jmairboeck:
Make the ALL value upper case, as suggested by @waddlesplash.

This is a breaking change! The recipe format version should be bumped when merging this (ideally together with #349).

The other case-normalized attributes types are:
 * Provides and Requires List (lower case)
 * YesNo (lower case)
 * Phase (upper case)

Should these also be made case-sensitive? (probably only Phase)

### DIFF
--- a/HaikuPorter/ConfigParser.py
+++ b/HaikuPorter/ConfigParser.py
@@ -164,7 +164,7 @@ class ConfigParser(object):
 			elif attrType == MachineArchitecture:
 				entries[key] = {}
 				knownArchitectures = MachineArchitecture.getAll()
-				valueString = valueString.lower()
+				#valueString = valueString.lower()
 				if valueString not in knownArchitectures:
 					architectures = ','.join(knownArchitectures)
 					sysExit('%s refers to unknown machine-architecture %s\n'
@@ -173,7 +173,7 @@ class ConfigParser(object):
 				entries[key] = valueString
 			elif attrType == Architectures:
 				entries[key] = {}
-				for value in [v.lower() for v in valueString.split()]:
+				for value in valueString.split():
 					architecture = ''
 					if value.startswith('?'):
 						status = Status.UNTESTED
@@ -185,13 +185,14 @@ class ConfigParser(object):
 						status = Status.STABLE
 						architecture = value
 					knownArchitectures = Architectures.getAll()
-					if architecture == 'all':
+					if architecture == 'ALL':
 						if len(entries[key]) > 0:
-							sysExit("%s specifies 'all' after other architectures"
+							sysExit("%s specifies 'ALL' after other architectures"
 									% (filename))
 						for machineArch in MachineArchitecture.getAll():
 							entries[key][machineArch] = status
 					else:
+						#architecture = architecture.lower()
 						if architecture not in knownArchitectures:
 							architectures = ','.join(knownArchitectures)
 							sysExit('%s refers to unknown architecture %s\n'


pulkomandy:
> This is also the reason why "ALL" must come first. If you expand it like that, this is easy to see.

I think this is a sign that the implementation is a bit hacky. Using uppercase to make it more obvious that it is a hack with counter-intuitive usage is not great.

Maybe something like `ARCHITECTURES = ALL_EXCEPT("x86_gcc2")` would make more sense. Then it can be documented (and even implemented) as an actual function or macro and not a string substitution hack.

And we can also have a predefined constant that allows to write `ARCHITECTURES = ALL` for the simplest cases.

Begasus:
Like "x86_64 ALL" would only apply on x86_64 and not the others?
EDIT: if so, we already to those with "?all x86_64" ?

jmairboeck:
No, I meant that ALL doesn't affect x86_64 when specified after it, so that the order doesn't matter again.

pulkomandy:
My problem is it is used inside a string which is quite confusing, I would say macros usually do not work like that in other languages.

I would prefer either for it to be outside the string (as I proposed) or maybe as a variable ($ALL or some other prefix).

The idea being to keep the recipe syntax close to bash, and not add extra "magic" on top of that where not needed.

That being said, I don't care enough about this to provide a different implementation, so if the current solution seems good enough, you can go with it.

Begasus:
Don't have any strong opinion here, but please don't over-complicate things just for the sake of it.

jmairboeck:
> > [...] maybe as a variable ($ALL or some other prefix).
> 
> Unfortunately this doesn't work so easily, because we also need to apply the prefix character to all of them. If $ALL just expands to all architectures, "?$ALL" would only put the ? before the first one.

Another idea: we could make the prefix character "sticky", so that it applies to all following values until it is changed without the need for it to be repeated, and add a new one (like `=`) to denote stable. Then we could just make a normal variable out of it. But that would be quite a big breaking change.

Example:

```
?x86_64 riscv64 =x86_gcc2
```
would be what is now written as
```
?x86_64 ?riscv64 x86_gcc2
```

pulkomandy:
> It's functionally a "macro" which token-pastes all architectures. We generally define macros in uppercase in Haiku coding style, so what's the problem here?

So, trying to think about it at a higher level:

The fact that "all" is implemented as a macro should be an implementation detail. At the moment, it isn't, because the way the macro works is not fully transparent, and can result in unexpected behavior. Specifically, its position in the list makes it do different things. For example "!x86_gcc2 x86_64" is the same as "x86_64 !x86_gcc2", but "!x86_gcc2 all" is not the same as "all !x86_gcc2".

I think (but I can only guess, since no good explanation was given) that this is the original problem leading to this discussion in the first place.

In that context, there are many ways to solve this problem:

- Make "all" behave "as expected", that is, instead of expanding to all architectures, it should expand to "all architectures not otherwise specified in the string". Maybe it should be renamed to "others" then, and you could do something like "!x86_gcc2 others" and it would do what makes sense: mark as broken on x86_gcc2 and working on all other architectures.
- Keep the current behavior, which is unintuitive, and try to make it more obvious that "all" has this unintuitive behavior by making it stand out. This change suggests using uppercase, but I find that not really helpful. It makes it stand out and look different, but I wouldn't think of macro expansion when looking at it inside a string. The chain of thought to make that work is something like: "This is in uppercase unlike the others, must be something special" -> "Macros in C in Haiku sourcecode also use uppercase, this might be a macro" -> "I still have to look up the documentation or sourcecode of haikuporter to understand what it does" -> "Ok, this is a macro that expands to all architectures while optinally prepending a prefix to all of them" -> "Now I can run that algorithm in my head, and then I have to know that it's possible to override the status for an architecture by putting it again later in the string, but not by putting it before ALL". I don't think this is very helpful, the solution is very oriented to knowing the implementation and the coding style of an unrelated programming language ("uppercase is for macros" is a C++ thing).
- Introduce a whole new syntax like the one I proposed, or any other one that fits. I would be happy with something like $ALL (which makes it obvious that something like bash expansion rules would apply, something we routinely use in haikuports recipes), but it's not exactly obvious how !$ALL would expand then. I would also be fine with introducing other markers that make this explicitly look like a macro, preferably drawing from bash and Python rather than C++ conventions (especially when used in a way that wouldn't quite work in C++, where macros inside strings are pretty unusual).
- Another option is to have haikuporter simply catch "nonsensical" settings of the ARCHITECTURES variable before doing the "all" expansion. For example, prevent "all" from being used anywhere but as the first item in the ARCHITECTURES list. Then we can have an error message with all the explanations we want in there.

jmairboeck:
> * Make "all" behave "as expected", that is, instead of expanding to all architectures, it should expand to "all architectures not otherwise specified in the string". Maybe it should be renamed to "others" then, and you could do something like "!x86_gcc2 others" and it would do what makes sense: mark as broken on x86_gcc2 and working on all other architectures.

See #366 for that (still called "all" though).

> * Another option is to have haikuporter simply catch "nonsensical" settings of the ARCHITECTURES variable before doing the "all" expansion. For example, prevent "all" from being used anywhere but as the first item in the ARCHITECTURES list. Then we can have an error message with all the explanations we want in there.

This is what it currently does (since #356).

PR #359, opened 2026-01-05
pyproject.toml: add script declaration to install haikuporter

### LINKED ISSUE #277: pip3 install no longer adds executable to path
After the changes @jurgenwigg  put in place to modernize haikuporter, ```pip3 install``` no longer places haikuporter or haikuporter.py into a PATH.   From what I can tell, it no longer even installs haikuporter.py anywhere.

jmairboeck:
Apparently, poetry does support installing existing raw files as scripts. See https://python-poetry.org/docs/pyproject/#scripts-1

This should fix #277.

----

I do agree with @OscarL that poetry has too many dependencies, which we probably will not need. Should we maybe switch the build backend back to setuptools? I would keep `pyproject.toml` though. Setuptools seems to also still support raw scripts, although it is discouraged. See https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration (the `script-files` option).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,9 @@ mkdocs = "^1.4.2"
 mkdocs-material = "^8.5.10"
 mkdocstrings = {extras = ["python"], version = "^0.19.0"}
 
+[tool.poetry.scripts]
+haikuporter = {reference = "haikuporter", type = "file"}
+
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"


OscarL:
LGTM (but havent locally tested).

----

I would very much like to get rid of all this poetry stuff. If we're gonna switch, I'd like to use the leanest possible build system.

`pyproject.toml` is pretty much the modern standard, so yes, that should stay.

I have no experience creating Python packages myself, but based on my experience over HaikuPorts... I'm partial to `flit-core`, and using `build` + `install`.

As a quick test, I've replaced HaikuPorter's `pyproject.toml` with the following:

```toml
[build-system]
requires = ["flit_core"]
build-backend = "flit_core.buildapi"

[project]
name = "HaikuPorter"
authors = [ { name = "Haiku Inc.", email = "contact@haiku-inc.org" } ]
dynamic = [ "version" ]
description = "The tool that builds HaikuPorts recipes."

[project.scripts]
haikuporter = "HaikuPorter.Main:main"
```

And applied the following diff to `HaikuPorter/Main.py`:

```diff
diff --git i/HaikuPorter/Main.py w/HaikuPorter/Main.py
index c597d30..bb44653 100644
--- i/HaikuPorter/Main.py
+++ w/HaikuPorter/Main.py
@@ -12,6 +12,7 @@
 
 # -- Modules ------------------------------------------------------------------
 
+import logging
 import os
 import re
 import sys
@@ -21,7 +22,7 @@ from subprocess import check_call
 from .BuildPlatform import buildPlatform
 from .Configuration import Configuration
 from .DependencyAnalyzer import DependencyAnalyzer
-from .Options import getOption
+from .Options import getOption, parseOptions
 from .PackageRepository import PackageRepository
 from .Policy import Policy
 from .RecipeAttributes import getRecipeFormatVersion
@@ -958,3 +959,15 @@ class Main(object):
 				port = allPorts[portID]
 				print('%s	[%s]' % (portID, port.category))
 				port.checkPortReleases()
+
+
+def main():
+	logger = logging.getLogger("buildLogger")
+	logger.setLevel(logging.INFO)
+	logger.addHandler(logging.StreamHandler())
+
+	Main(*parseOptions())
+
+
+if __name__ == "__main__":
+	main()
```

Then I was able to use `python3 -m build --wheel --no-isolation` to create the .whl file, and install it with `pip3 install --user dist/haikuporter-1.3.3-py2.py3-none-any.whl`, which properly added a functional `haikuporter` script into `~/config/non-packaged/bin`.

----

With the change in Main.py, we could then simplify haikuporter.py to be just:

```python
#!/usr/bin/env python3

from HaikuPorter.Main import main

main()
```

jmairboeck:
I'm also fine with flit-core. I suggested setuptools because flit-core doesn't support raw scripts as entry points, AFAIK, but I think your `Main.py` changes should work too.

jmairboeck:
I'll merge this now and do the refactoring to flit-core separately.

PR #360, opened 2026-01-06
switch build system to flit-core, clean up dependencies

jmairboeck:
* Remove codecov and pytest-cov
 * Remove tox.ini
 * Cleanup .gitignore (mostly restore the version before https://github.com/haikuports/haikuporter/pull/259)


pyproject.toml
```diff
@@ -1,106 +1,21 @@
...
-[tool.coverage.report]
-skip_empty = true
+[project.scripts]
+haikuporter = "HaikuPorter.Main:main"
 
-[tool.coverage.run]
-branch = true
-source = ["HaikuPorter"]
+[tool.flit.external-data]
+directory = "data"
```
jmairboeck: I'm not sure about this one. This includes the haikuports sample config in the wheel and should automatically install it correctly when using `installer` in the recipe. But I'm not sure if this would be problematic for the buildmaster. (`pip3 install` installs it in `non-packaged/settings/haikuports.conf`)

pyproject.toml
```diff
@@ -1,106 +1,21 @@
...
+dynamic = [ "version" ]
+description = "The tool that builds HaikuPorts recipes."
+requires-python = ">=3.10"
 
-[tool.ruff.per-file-ignores]
-"tests/*" = ["S101"]
+[project.optional-dependencies]
+test = [ "pytest" ]
+doc = [ "sphinx" ]
+buildmaster = [ "boto3", "paramiko", "pymongo" ]
```
jmairboeck: @kallisti5 this should theoretically allow to use `pip3 install /tmp/haikuporter[buildmaster]` in the Dockerfile. (untested)

haikuporter.py
```diff
@@ -1,15 +1,8 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# Copyright 2013-2014 Oliver Tappe
+# Copyright 2026 Oscar Lesta.
```
OscarL: Feels odd for me to claim credits for this two-liner :-D

Do we have a "HaikuPorter contributors", we could use?

pyproject.toml
OscarL: I assume you locally created a  `data` dir and moved/copied haikuports.conf in there, right? (I just get an error about `data` key not being a dir).

Just for kicks, I've tried with `directory = "buildmaster"`, it does adds the full dir into the .whl file (including .gitignore files, and such). On installation, all the files and folders under `buildmaster/` ends up under `~/config/non-packaged/` indeed.

waddlesplash: > But I'm not sure if this would be problematic for the buildmaster.

The buildmaster just uses a git checkout.

jmairboeck: Yes, but if the Dockerfile is switched back to `pip3 install`, would the config file in be a problem there?

PR #362, opened 2026-01-24
support not creating a base package

### LINKED ISSUE #192: Python base packages don't provide anything ...

jmairboeck:
This is implemented by making PROVIDES optional and making the recipe not declare anything to provide. PROVIDES de facto defines (sub)packages, so leaving it out should be natural for a package that isn't actually created.

~~A build package (with fake PROVIDES) is still created to have a writable $prefix.~~

A policy check is added that the main packaging directory is empty after INSTALL() and that the recipe still at least declares one actual package.

Fixes #192.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -88,6 +88,20 @@ class Policy(object):
 		if self.strict and self.violationEncountered:
 			sysExit("packaging policy violation(s) in strict mode")
 
+	def checkPort(self):
+		self.violationEncountered = False
+
+		self._checkDummyPrefixIsEmpty()
+
+		if self.strict and self.violationEncountered:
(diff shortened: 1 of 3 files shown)


jmairboeck:
Any more comments on this? Can we merge this (and get it deployed)?

This would also be useful for the noto_*_cjk font packages, in addition to python.

OscarL:
Just did a quick test (by commenting out PROVIDES in the `dev-python/wheel` recipe. Worked as advertized. Nice!

+1 from me. Thanks @jmairboeck!

PR #365, opened 2026-02-01
ShellScriptlets: don't check whether package prefixes exist

jmairboeck:
The check doesn't work outside the chroot (it would require the package in question to be installed locally), so it caused many warnings when getPackagePrefix is used in e.g. defineDebugInfoPackage.

Instead of checking the existence and probing a second time using $portBaseName, check whether the package is an "any" arch package and use $portBaseName in the first place. Note that when the base package is "any", no check is necessary because $portName and $portBaseName are the same in this case anyway.


Begasus:
Not really sure how to trigger this?

Issue #339, opened 2025-08-08
`"all !x86_gcc2"` vs `"!x86_gcc2 all"` in `ARCHITECTURES`


OscarL:
A funny one:

Was trying to understand why the [repo_consistency.txt](https://build.haiku-os.org/buildmaster/master/x86_gcc2/repo_consistency.txt) was reporting issues with `ceres_solver` (which is `"!x86_gcc2 all"`, so it shouldn't even be there), noticed the following:

On beta5, 32 bits, latest HaikuPorter code:

`hp ceres_solver` starts downloading the tarball instead of stopping as the recipe is marked broken/disabled for gcc2.

Changed the recipe to `"ARCHITECTURES="all !x86_gcc2"`, and then I got:

```
Warning: skipping ceres_solver-2.0.1~git, as it is broken on the target architecture.
Error: No version of ceres_solver can be built
```

as expected.

Haven't looked at HaikuPorter code yet, filing this one before I forget.

----

Side note...

`inrecipe '"all !x86_gcc2"' | wc -l` reports 2001 recipes, while `inrecipe '"!x86_gcc2 all"' | wc -l` gave only 4 results at the time I've checked, with some of those being only for embedded stuff. Might explain why we didn't noticed this one earlier?

waddlesplash:
`all` basically causes HaikuPorter to add all architectures. Probably it should be illegal to put it anywhere except first, that should be an error.

PR #366, opened 2026-02-02
Allow the "all" architecture macro to appear anywhere

jmairboeck:
"all" now only sets architecture entries that haven't been set explicitly before yet. This allows it to occur anywhere in the list again. Architectures which come before "all" are not overwritten by it, but those coming after it still override the defaults set by "all".

This is a better fix for #339.

----

This conflicts with #357 and still uses "all" in lower case, but I wanted to keep these two concerns separately.


jmairboeck:
> Having it be required to appear first, if at all, may make sense, though?

This is what the current implementation does (after #356). This PR is an alternative suggestion. Generally, the order of the architecture values doesn't matter (conceptually), and this implements this for "ALL" too.

Issue #363, opened 2026-01-25
Why are ADDITIONAL_FILES indexable and [BUILD|TEST]_(PRE)REQUIRES extendable?


jmairboeck:
`ADDITIONAL_FILES` are defined as indexable (i.e. per source) and are mostly handled at the source level.

`BUILD_REQUIRES`, `BUILD_PREREQUIRES` and `TEST_REQUIRES` are defined as extendable (i.e. per package).

All these attributes look logically more like port-level (i.e. global per recipe) attributes to me. Why were they defined like that? I don't think having e.g. `ADDITIONAL_FILES_2` or `BUILD_REQUIRES_devel` make any logical sense, and in fact none of these are used in haikuports.

For the three extendable ones, is it because they are used in the .DependencyInfo files?

For additional files, it looks like the directory they are in is named e.g. `additional-files-2` if the index is used (similar to `sources-2`).

jmairboeck:
I added 2 pull requests which move the attributes to global level:

- #369 for additional files
- #370 for the requires attributes

Everything still seems to work normally with both of them applied.

PR #369, opened 2026-03-08
move ADDITIONAL_FILES from Source to Port level

jmairboeck:
It doesn't really make sense to have additional files per source and separate "additional-files-2" etc. directories. So handle these completely globally per port instead.

Part of #363.

### DIFF
--- a/HaikuPorter/Port.py
+++ b/HaikuPorter/Port.py
@@ -623,6 +623,18 @@ class Port(object):
 			if source.referencesFiles(files):
 				return True
 
+		if self.additionalFiles:
+			for additionalFile in self.additionalFiles:
+				if os.path.isdir(additionalFile):
+					# ensure there is a path separator at the end
+					additionalFile = os.path.join(additionalFile, '')
+					for fileName in files:
+						if os.path.commonprefix([additionalFile, fileName]) \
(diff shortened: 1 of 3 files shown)


HaikuPorter/Port.py
```diff
@@ -1405,6 +1437,10 @@ def _adjustToChroot(self):
 		self.packagingBaseDir = self.packagingBaseDir[pathLengthToCut:]
 		self.packageInfoDir = self.packageInfoDir[pathLengthToCut:]
 		self.hpkgDir = self.hpkgDir[pathLengthToCut:]
+		self.additionalFiles = [
+			additionalFile[pathLengthToCut:]
+			for additionalFile in self.additionalFiles
+		]
```
jmairboeck: Is this actually correct? I think this would remove length of the workdir path from the addional-files paths, but the path should be actually the additional-files directory instead, if I see it correctly.

It seems that this is actually never used inside the chroot, which is why it doesn't seem to matter much for now. The existing code (which is now deleted from Source.py, see there) has the same bug though.

jmairboeck: I fixed this now by only cutting the base dir for additional files.

PR #376, opened 2026-05-30
Policy: check that all provided cmd:, lib: and devel: entries exist

jmairboeck:
This contains also some minor cleanups.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -125,6 +125,8 @@ class Policy(object):
 			self._violation('no matching self provides for "%s"'
 				% self.package.name)
 
+		foundEntries = set()
+
 		# everything in bin/ must be declared as cmd:*
 		binDir = os.path.join(self.package.packagingDir, 'bin')
 		if os.path.exists(binDir):
@@ -133,9 +135,11 @@ class Policy(object):
 				if entry == self.package.secondaryArchitecture:
 					continue
 				name = self._normalizeResolvableName('cmd:' + entry)
-				if name.lower() not in self.provides:
-					self._violation('no matching provides "%s" for "%s"'
-						% (name, 'bin/' + entry))
+				if name not in self.provides:
+					self._violation('no matching provides "%s" for "bin/%s"'
+						% (name, entry))
+				else:
+					foundEntries.add(name)
 
 		# library entries in lib[/<arch>] must be declared as lib:*[_<arch>]
 		libDir = os.path.join(self.package.packagingDir,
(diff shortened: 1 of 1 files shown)


HaikuPorter/Policy.py
```diff
@@ -166,9 +172,22 @@ def _checkProvides(self):
...
+					foundEntries.add(name)
+
+		# all cmd:, lib: and devel: entries must exist
+		for entry in self.provides:
+			if entry.startswith('cmd:') and entry not in foundEntries:
+				self._violation('provides "%s" doesn\'t exist in "bin"' % entry)
+			if entry.startswith('lib:') and entry not in foundEntries:
+				self._violation('provides "%s" doesn\'t exist in "lib%s"'
+					% (entry, self.secondaryArchSubDir))
+			if entry.startswith('devel:') and entry not in foundEntries:
```
jmairboeck: We use `devel:` also for some header-only libraries. We could use `devel:lib` (and `lib:lib`) to check here to (probably) not catch these.

HaikuPorter/Policy.py
```diff
@@ -166,9 +172,22 @@ def _checkProvides(self):
...
+					self._violation('no matching provides "%s" for "develop/lib%s/%s"'
+						% (name, self.secondaryArchSubDir, entry))
+				else:
+					foundEntries.add(name)
+
+		# all cmd:, lib: and devel: entries must exist
+		for entry in self.provides:
+			if entry.startswith('cmd:') and entry not in foundEntries:
+				self._violation('provides "%s" doesn\'t exist in "bin"' % entry)
+			if entry.startswith('lib:') and entry not in foundEntries:
```
jmairboeck: There is at least one known "false positive" for this one: perl provides `lib:libperl`, but that is not in `$libDir`, but in `lib/perl/...` (so not even a subdirectory of `$libDir` on x86, so I don't think it's feasible to find it there). Therefore, it also doesn't have a corresponding `devel:` entry. The provides entry is necessary because other packages link to it and thus require it.

Do we want to live with that policy violation in this case?

HaikuPorter/Policy.py
jmairboeck: I changed this, so that it should now catch only actual libraries.

jmairboeck:
I added library version checks as well, in a separate commit.

Note: The check is not perfect. As each file is checked individually, this usually results in two policy messages if the version specification is not correct, similar to this:

```
Warning: POLICY ERROR: version of provides "lib:libfoo" doesn't match library (expected "1")
Warning: POLICY ERROR: version of provides "lib:libfoo" doesn't match library (expected "1.2.3")
```

This means that the correct version specification for "lib:libfoo" should be "= 1.2.3 compat >= 1".

HaikuPorter/Policy.py
waddlesplash: Probably yes.

waddlesplash: I guess it will have to be a warning only and not an error then?

waddlesplash:
I think we should enhance the library version checks to be able to fully build the expected version/compat specification, and insert it if not present already, and error if it's present but doesn't match. If you aren't sure how to do this I can take a look.

HaikuPorter/Policy.py
jmairboeck: Yes, all policy violations are warnings by default.

jmairboeck:
Thanks! I'll split the second commit into its own branch in the upstream repo, so that you (or others) can work more easily on it and merge the first part of this one.

PR #377, opened 2026-06-04
Policy: add version checks for library provides and automatically add missing versions

jmairboeck:
Each .so file with a version suffix in $libDir and $developLibDir is checked whether that version suffix matches the corresponding provides entry, i.e. it is one of the declared versions (either the full one or the compat version).

For this, the result of _parseResolvableExpressionList is changed from a set to a dict. The names are still used as keys and the value is a list of versions that have been found.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -127,7 +127,7 @@ class Policy(object):
 			self._violation('no matching self provides for "%s"'
 				% self.package.name)
 
-		foundEntries = set()
+		cmdEntries = set()
 
 		# everything in bin/ must be declared as cmd:*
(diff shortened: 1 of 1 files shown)


jmairboeck:
I think I got it now, that the versions are checked correctly and only one violation message is printed for each entry.

jmairboeck:
I implemented automatically adding missing version numbers to PROVIDES.

The code is not very pretty and can probably be improved (I'm still not very good at python).

Also, while testing this I noticed that it is maybe not that useful after all, because we still need the `libVersion` in the recipe for the debuginfo filenames. `libVersionCompat` can be removed though.

Issue #380, opened 2026-06-08
DependencyInfos of ports without a main package are always updated


jmairboeck:
This is because the "main dependency info" is checked for whether the dependency infos are outdated, but that doesn't exist in this case because there is no package for it, and so they are always considered outdated.

Using another package's dependency info file as the "main" one is probably the most logical solution here, but that would require parsing the recipe to get the name of that.

jmairboeck:
Another possibility maybe would be to create a dummy dependency info file, but I'm not sure what consequences that would have. I would expect at least some warnings because of an invalid dependency info file.

OscarL:
Ah, I see you already beat me to it (had written my "bug report" in a .txt, before getting back online).

Only thing for me to add would be: besides the "updating dependency infos of...", the "Looking for stale cached recipes .." step also gets called for each recipe without a base/main package.

jmairboeck:
> Only thing for me to add would be: besides the "updating dependency infos of...", the "Looking for stale cached recipes .." step also gets called for each recipe without a base/main package.

Yes, for the same reason.

jmairboeck:
I think we will need a separate timestamp marker file for this.

PR #381, opened 2026-06-08
Automatically create and remove the dummy-prefix directory if one is needed.

jmairboeck:


HaikuPorter/Port.py
```diff
@@ -1297,6 +1297,10 @@ def _recreatePackageDirectories(self):
 			self.packageInfoDir, self.packagingBaseDir,
 			self.buildPackageDir, self.hpkgDir
 		]
+		# if we don't have a base package, create the dummy-prefix directory
+		if ('PROVIDES' not in self.recipeKeysByExtension['']
+			or len(self.recipeKeysByExtension['']['PROVIDES']) == 0):
```
jmairboeck: Should we maybe do this always instead, to better support switching between having a main package and not? Or is that too niche of a use case?

OscarL: > to better support switching between having a main package and not

I'd say it's just too niche to bother (I was only attempting that back-and-forth, while failing to see/fix one of my many silly mistakes :-P).

OscarL: I guess making sure `directoriesToCreate.append(self.dummyPrefixDir)` if no PROVIDES is fine (albeit I don't think I saw a problem with that dir *not* being created, I just failed to leave it empty).

The check for present-but-empty PROVIDES might be "overkill". I would expect an empty PROVIDES to still create/use the regular base/main package dirs instead.

OTOH, cleaning up the `dummy-prefix` when build was OK on `ShellScriplets.py`, seems sensible.

OscarL:
I guess that, besides #380, this also was prompted by my ramblings over IRC :-)

To be honest... the issues I was having where the `dummy-prefix` dir got involved, was a "I'm just still unfamiliar with this particular policy error" type of situation.

I was getting:

`POLICY ERROR: main package isn't defined but not all files have been moved to other packages"`

and failing to properly remove the offending empty dir. Finally figured how to "fix" it (after a few attempts, because I'm dumb).

I think it *might* have been a little less confused if the policy error read "but not all dirs or files" instead of just "files", and if it also printed the offending dir(s)/file(s) in question.

But I still consider this one as a "skill-issue" on my part.

jmairboeck:
The missing part here was it not being part of the `directoriesToRemove`, and then I thought we might as well create it automatically like the other directories, at least for consistency.

HaikuPorter/Port.py
jmairboeck: I changed this to always create and clean up the dummyPrefixDir regardless. I think this should not be a problem.

PR #383, opened 2026-06-28
cargo-to-recipe: try to fix SOURCE_URIs of crates ending in numbers

jmairboeck:
Remove the version number pattern only once from the cargo name, instead of potentially multiple times.

This should fix cases like "md-5" being shortened to just "md" because the end of it looks like a version number.

### DIFF
--- a/tools/cargo-to-recipe.sh
+++ b/tools/cargo-to-recipe.sh
@@ -155,7 +155,6 @@ info=$(
 		sed -e '0,/\[metadata\]/d
 			s/checksum //
 			s/(.*)//
-			s/ /-/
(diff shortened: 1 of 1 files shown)


jmairboeck:
This quick fix will probably not work correctly, as I see version numbers like "1.1.2+spec-1.1.0" which contain `-` themselves.

Unfortunately I don't understand the awk code enough that parses the Cargo.lock file, to fix this properly ...

jmairboeck:
I think I got it now.

The `[metadata]` case is not tested, but as far as I can tell, this should work too.

Good to merge?

PR #387, opened 2026-07-24
Fix errors for tar files when extracting

Begasus:
appstream for one: 'appstream-1.1.3/tests/samples/compose/usr/share/metainfo/org.example.nonexistent-badlink.metainfo.xml' is a link to an absolute path


OscarL:
Relevant docs, for reference:

- https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall
- https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter
- https://peps.python.org/pep-0706/

This change should theoretically be safer than it was before with 3.10.

PR #388, opened 2026-07-25
Policy: check that all required libraries are actually used

### LINKED ISSUE #386: Warn if REQUIRES declared that aren't actually required by any library
This could happen legitimately in the case of runtime dependencies that aren't linked directly, but that will be a rare case and we can just add comments in those recipes to ignore the warning.

jmairboeck:
Fixes #386.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -199,6 +199,8 @@ class Policy(object):
 		if not isCommandAvailable('readelf'):
 			return
 
+		self.foundUsedRequires = set()
+
 		# check all files in bin/, apps/ and lib[/<arch>]
 		for directory in ['bin', 'apps', 'lib' + self.secondaryArchSubDir]:
 			dir = os.path.join(self.package.packagingDir, directory)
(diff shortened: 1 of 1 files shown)


jmairboeck:
I think this check is actually too strict, mostly because the complement doesn't check all directories where binaries may be installed. So this now triggers for all python and perl packages, which include native code and have dependencies, like binding packages, but this check doesn't get that because it doesn't check libraries in deeper subdirectories.

I just tested xml_libxml (the perl bindings for libxml2) and got a policy warning for it.

There are probably many more examples, like translators, which are installed in addons, which isn't checked.

Should we maybe revert this?

waddlesplash:
> mostly because the complement doesn't check all directories where binaries may be installed

That sounds wrong, and could be a source of problems in the opposite direction, if there are binaries in those directories with library requirements that are not declared in the package. Shouldn't we rather fix that bug?

jmairboeck:
So we need to check all subdirectories of lib, apps, addons etc.? What else?

waddlesplash:
Probably we need to check everything for binaries except `develop`.

jmairboeck:
I found a counter example even for that: gcc is installed in develop/tools.

Begasus:
Things can go deeper also: /boot/system/data/Qt6/qml/org/kde/i18n/localeData

PR #390, opened 2026-08-01
Policy: check for missing libraries recursively

### LINKED ISSUE #389: Improve "missing library" policy checks
The "missing library" policy check is currently only checking a few selected directories (bin, apps and lib$secondaryArchSubDir) and only one or two levels deep. It therefore misses all binaries installed in other directories or in deeper subdirectories.

As a consequence, the new redundancy check introduced in #388 causes many false positives (it marks everything that has not been found used by the main check before).

The main check should therefore be extended to also check other locations where binaries may be installed, and arbitrarily deep.

jmairboeck:
Also, check nearly all directories, not just bin, apps and lib[/x86]. Only 'data' and 'documentation' are skipped, because they shouldn't contain any binaries. Symlinks are also skipped now. (This is needed for devel packages now.)

This way, we also find binaries installed in other directories or deeper subdirectories. Fixes #389.

Remove the check for readelf, as that is always available (it is part of the scriptlet prerequires). Instead, skip "any" arch packages.

### DIFF
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -202,13 +202,6 @@ class Policy(object):
 		self.foundUsedRequires = set()
 
 		for dir, dirs, files in os.walk(self.package.packagingDir):
-			# don't check some top-level directories
-			if dir == self.package.packagingDir:
-				if 'data' in dirs:
-					dirs.remove('data')
-				if 'documentation' in dirs:
-					dirs.remove('documentation')
-
 			for entry in files:
 				path = os.path.join(dir, entry)
 				if not os.path.islink(path):
--- a/HaikuPorter/Policy.py
+++ b/HaikuPorter/Policy.py
@@ -12,6 +12,7 @@ from subprocess import check_output
 
 from .ConfigParser import ConfigParser
 from .Configuration import Configuration
+from .RecipeTypes import Architectures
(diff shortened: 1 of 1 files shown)


HaikuPorter/Policy.py
```diff
@@ -194,28 +195,24 @@ def _normalizeResolvableName(self, name):
...
-			dir = os.path.join(self.package.packagingDir, directory)
-			if not os.path.exists(dir):
-				continue
+		for dir, dirs, files in os.walk(self.package.packagingDir):
+			# don't check some top-level directories
+			if dir == self.package.packagingDir:
+				if 'data' in dirs:
+					dirs.remove('data')
+				if 'documentation' in dirs:
+					dirs.remove('documentation')
```
jmairboeck: This excludes the top-level `data` and `documentation` directories as an optimization. If we should check everything instead, we could remove this.

waddlesplash: How long does the check take to run? If it's not too slow I'd just not exclude anything.

jmairboeck: I only tested rather small packages for now, and it wasn't very significant. I guess we can remove it for now and add it back if it turns out too slow.

@Begasus how was your experience so far?

PR #394, opened 2026-08-19
switch build to setuptools, introduce setuptools-scm

jmairboeck:
This gets rid of specifying the version number manually in `__version__.py` and instead reads it from git tags.

flit-core doesn't support this natively, thus we switch to setuptools for the build (which is required by setuptools-scm anyway).

Switch the buildmaster docker build back to using pip.

### DIFF
--- a/buildmaster/backend/Dockerfile
+++ b/buildmaster/backend/Dockerfile
@@ -19,6 +19,7 @@ RUN apt-get update \
 	&& apt-get -y install attr autoconf automake bison coreutils curl flex \
 		gawk gcc g++ git libcurl4-openssl-dev make nasm \
 		python3 python3-isort python3-boto3 python3-pymongo python3-paramiko \
+		python3-setuptools-scm python3-pip \
 		tar texinfo vim wget zlib1g-dev; \
 	if [ "$(uname -m)" = "x86_64" ]; then apt install -y gcc-multilib; fi; \
 	apt-get clean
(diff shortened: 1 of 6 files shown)


jmairboeck:
I still need to check the Docker build, thus this is still a draft.

buildmaster/backend/Dockerfile
```diff
@@ -33,10 +34,8 @@ ADD . /tmp/haikuporter
 COPY --from=host-tools /tmp/haiku/generated/objects/linux/*/release/tools/package/package /usr/local/bin/
 COPY --from=host-tools /tmp/haiku/generated/objects/linux/*/release/tools/package_repo/package_repo /usr/local/bin/
 
-RUN PYVERSION=$(python3 -V | awk '{ print $2 }') \
-	&& mkdir -p /usr/local/lib/python${PYVERSION%.*}/dist-packages \
-	&& mv /tmp/haikuporter/HaikuPorter /usr/local/lib/python${PYVERSION%.*}/dist-packages/ \
-	&& mv /tmp/haikuporter/haikuporter.py /usr/local/bin/haikuporter \
+RUN --mount=source=.git,target=/tmp/haikuporter/.git,type=bind,rw \
```
jmairboeck, line 37: This does not work, it fails at `rm -rf /tmp` below because it can't remove the bind mount.

So we'll need to inject the version here anyway (via SETUPTOOLS_SCM_PRETEND_VERSION).

So, neither the haikuports recipe nor the docker build can use this directly, which means that doing it "manually" (e.g. via sed prior to the build) is probably just as good and we can keep using flit-core instead of setuptools.

PR #397, opened 2026-08-24
optionally enable the network again

jmairboeck:
This adds a new configuration setting ALLOW_NETWORK_IN_CHROOT to enable copying the network settings into the chroot again, either in general or for tests only. The network is still disabled by default.

### DIFF
--- a/HaikuPorter/ConfigParser.py
+++ b/HaikuPorter/ConfigParser.py
@@ -8,7 +8,7 @@
 import functools
 from subprocess import CalledProcessError, check_output
 
-from .RecipeTypes import (Architectures, Extendable, LinesOfText,
+from .RecipeTypes import (AllowNetworkMode, Architectures, Extendable, LinesOfText,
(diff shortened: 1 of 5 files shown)


waddlesplash:
What is this necessary for? IMHO we simply should not allow network in the chroot and never should have done so. It breaks having the SOURCES be the only source of data inside the chroot.

Begasus:
For packagers this could be handy, in some cases (mostly in TEST()) one needs additional python modules, having access to network enables to install/uninstall these modules then.
Those should only be done on local machines though (as I think this takes care of that as the user needs to manually enable it).

jmairboeck:
This is intended to enable it for local testing if needed. We also have options ALLOW_UNTESTED and ALLOW_UNSAFE_SOURCES, which go in a similar direction (disabled by default but can be enabled to allow certain features).

waddlesplash:
OK, I see.

PR #398, opened 2026-09-09
Compute full build order for bootstrap

przemub:
Previously only ports with circular deps were ordered. Others were built without dependency resolution, deadlocking very often (freetype before libpng16, gd before giflib).


waddlesplash:
CC @mmlr @jmairboeck

